PR #1, opened 2017-01-20
Fix golint linting errors

jsageryd:
Hi, I don't know if you accept pull-requests, but I went ahead and fixed all errors reported by [golint](https://github.com/golang/lint), except the ones about documenting what is exported. This should make it a little bit more idiomatic. Probably want to change e.g. the log level consts from `instana.Error` to something that more clearly indicates that it is a log level. But this is a start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ in your main function. The init function takes an `Options` object with the foll
 
 * **Service** - global service name that will be used to identify the program in the Instana backend
 * **AgentHost**, **AgentPort** - default to localhost:42699, set the coordinates of the Instana proxy agent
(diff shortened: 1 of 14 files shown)


pavlobaron:
thanks a lot @jsageryd ! Sure, pull requests are very welcome :) I did ignore these linter messages in the first post-linter fixes, focusing on most critical ones. Let me test and check with the current users before I merge. The hint with the log level sure makes sense.

pavlobaron:
sorry, it took some time to get some time and to check with others. lgtm, and happily compiles, thanks a lot!

PR #5, opened 2017-02-10
Gateway discovery without dependencies

mier85:
Just some refactoring and more importantly i replaced the call to local tools with some native go code.

### DIFF
--- a/fsm.go
+++ b/fsm.go
@@ -2,10 +2,10 @@ package instana
 
 import (
 	"os"
-	"os/exec"
 	"strconv"
 	"time"
 
+	"github.com/instana/golang-sensor/gateway"
 	f "github.com/looplab/fsm"
 )
 
@@ -55,13 +55,13 @@ func (r *fsmS) lookupAgentHost(e *f.Event) {
 			r.lookupSuccess(host)
 			return
 		}
-		gateway := r.getDefaultGateway()
-		if gateway == "" {
-			log.error("Default gateway not available. Scheduling retry")
+		gw, err := gateway.GetDefault()
+		if nil != err {
+			log.error("Default gateway not available. Scheduling retry", err)
 			r.scheduleRetry(e, r.lookupAgentHost)
 			return
 		}
-		go r.checkHost(gateway, func(b bool, host string) {
+		go r.checkHost(gw, func(b bool, host string) {
 			if b {
 				r.lookupSuccess(host)
 				return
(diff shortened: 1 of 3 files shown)


pavlobaron:
thanks! will look at it soon - currently preparing optimisations around FSM for more safety in communication with the agent.

pavlobaron:
@mier85 I just merged a PR of mine into master. Can you adjust this PR to it, then we can check and merge? Thanks!

mier85:
@pavlobaron I merged the master into the branch and there shouldn't be conflicts now.

pavlobaron:
@mier85 regarding default gateway: we've been thinking about that. Currently, all of our sensors in question use the same method parsing the table to find the gateway. The method you propose seems to utilise the fact that the go lib doesn't establish the actual connection. The problem with that is that other runtimes we have the sensors for might not do it that way, making us try to connect to a Google server which is critical. Parsing the table like we do is passive. So, currently we tend not do go with this go-specific solution. Looking forward to your opinion, maybe we're overlooking something here.

mier85:
@pavlobaron this lib is for go tools. So if there was ever a case that go is using a different connection for dialing out, than the default one, then you will have a  gateway determined which is not actually being used by the application and that would be a big issue. 

Other than  that, you will have external dependencies when running the application: ip, awk, cut and tr. It's not cross platform compatible and might even fail on different Linux systems. So to use the sensors you will have to adjust your deploy process and you need to figure out first that you would need to do that (the arp library I used doesn't yet support Windows yet, but that would actually also be easy to fix)

pavlobaron:
@mier85 ok, I'll look into some current issues on master. Need to talk to Ruby and Node guys about how we deal with golang sensor to go its own way with gateway, then will get back to you. Sorry I'm late on that, had a lot of work and travels.

pglombardo:
Hi @mier85 thanks for being patient on this PR.  

The `mostlygeek/arp` tool has a nice benefit of using the `/proc` table directly for linux but unfortunately ends up [shelling out](https://github.com/mostlygeek/arp/blob/master/arp_unix.go#L18) too for everything else which for our case would be Solaris, AIX, OSX etc...

The other item I noticed is the `net.Dial("udp", "8.8.8.8:80")` which in our case may not always be possible for security focused customers with private installations and strict firewalls.  (e.g. Banks etc.)

So I agree on the need for this PR but we may have to find a more versatile solution.  Any other ideas/objections?  Please let me know.

mier85:
Hi @pglombardo ,

Linux-boxes are (probably in at least 90%(according to my magic )) of the cases, what we will work with. So avoiding the shell in the most common use case is already a good step.

As for the privacy concern: as @pavlobaron pointed out, with the go-lib we are not actually establishing a connection at the point that net.Dial is called. So it is safe to use even with strict firewalls.

jw-s:
Hi Guys, Is there any progress on this PR? We use this sensor in our Go applications which run in scratch docker containers. At the moment we have to include the binaries which this library depends on. Would be great to have Go code to replace these exec calls.

pglombardo:
Hi @jw-s this has fallen under the radar.  I'll see if we can get this some attention soon.

BTW and for reference, we did this recently in our Python sensor by simply parsing /proc/net/route with no external dependencies: https://github.com/instana/python-sensor/pull/117

jw-s:
@pavlobaron Thanks for the reply. I'm happy to raise a PR by parsing `/proc/net/route` like the python implementation.

pglombardo:
@mier85 just a heads up that we merged #65 which removes the external dependencies.  Released and tagged in [v1.4.15](https://github.com/instana/go-sensor/releases/tag/v1.4.15).

Thanks all for the help and contributions!

PR #8, opened 2017-02-24
Span buffering

pavlobaron:
Spans are buffered according to the following strategy:

On the regular basis, spans will be buffered for a second up to the maximum of options.MaxBufferedSpans, defaulting to 1000. Whenever the span buffer is filled up to options.ForceTransmissionStartingAt (default 500) within this second, transmission will be made.

Leaving any of these options at zero causes default values to be used. Immediate spans transmission doesn't take place any more.


tracer.go
```diff
@@ -102,12 +102,13 @@ func NewTracer() ot.Tracer {
...
 // NewTracerWithOptions Get a new Tracer with the specified options.
 func NewTracerWithOptions(options *Options) ot.Tracer {
+	InitSensor(options)
+
 	return NewTracerWithEverything(options, NewRecorder())
 }
 
 // NewTracerWithEverything Get a new Tracer with the works.
 func NewTracerWithEverything(options *Options, recorder bt.SpanRecorder) ot.Tracer {
-	InitSensor(options)
```
pglombardo: @pavlobaron Not sure I understand the reason for this move.  Shouldn't `InitSensor` be called from the lowest level so we guarantee that metric collection is initialized regardless of the Tracer instantiation call?

pglombardo:
Merging and tagging for a release

PR #10, opened 2017-03-09
Use extended discovery/announce if there is a /proc filesystem

pglombardo:
Still a bit of testing remains
- [x] Initial implementation
- [x] Test on Ubuntu
- [x] Test on OSX

### DIFF
--- a/fsm.go
+++ b/fsm.go
@@ -1,6 +1,8 @@
 package instana
 
 import (
+	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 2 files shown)


fsm.go
```diff
@@ -127,17 +129,44 @@ func (r *fsmS) announceSensor(e *f.Event) {
 	log.debug("announcing sensor to the agent")
 
 	go func(cb func(b bool, from *FromS)) {
+
 		d := &Discovery{
 			PID:  os.Getpid(),
 			Name: os.Args[0],
 			Args: os.Args[1:]}
 
+		var socket net.Conn
```
bripkens: How do you make sure that this socket is used for the HTTP request? Looks like magic here :)

pglombardo: We're opening an additional dummy socket here before the HTTP announce is made.  We pass the fd/inode of this dummy socket in the announce and keep it open until the announce response is received.  

This is far simpler than modifying the announce payload before it's transmitted (but after socket is open).

So the cost is one extra socket connection on announce but I think it's fair given announce is only every 30 seconds or so.

bripkens: Right, sounds good!

PR #17, opened 2017-03-28
Fix error checking and add a recover

### LINKED ISSUE #16: net.Dial timeouts and connection errors are not handled correctly
If an error is returned by `net.Dial`, we need fix our handling:
https://github.com/instana/golang-sensor/blob/master/fsm.go#L140-L142

In case of an error, the instrumentation can panic:

```
March 28th 2017, 08:28:07.709   goroutine 1011375 [running]:
March 28th 2017, 08:28:07.709   
March 28th 2017, 08:28:07.709   panic(0x9d3ea0, 0xc420497580)
March 28th 2017, 08:28:07.709   panic: interface conversion: net.Conn is nil, not *net.TCPConn
```

cc: @codec

pglombardo:
This fixes #16 (net.Dial timeouts and connection errors are not handled correctly)

### DIFF
--- a/fsm.go
+++ b/fsm.go
@@ -138,21 +138,22 @@ func (r *fsmS) announceSensor(e *f.Event) {
 		d := &Discovery{PID: os.Getpid()}
 		d.Name, d.Args = getCommandLine()
 
-		var socket net.Conn
 		if _, err := os.Stat("/proc"); err == nil {
-			if socket, err := net.Dial("tcp", r.agent.host+":42699"); err == nil {
+			if addr, err := net.ResolveTCPAddr("tcp", r.agent.host+":42699"); err == nil {
+				if tcpConn, err := net.DialTCP("tcp", nil, addr); err == nil {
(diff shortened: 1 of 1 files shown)


mier85:
While it is a good idea to recover in case of panic, you can also check if a type conversion is successful by doing this: 
tcpConn, ok := socket.(*net.TCPConn)

ok is now a boolean which tells you whether the type conversion was successful, and then you can process from there.

mier85:
On a second note, when you use net.DialTCP instead of net.Dial then there is no need to even do the type conversion

pglombardo:
Excellent suggestions @mier85 - much appreciated.

I like how the `defer`/`recover()` catch all gives us peace of mind in case of unexpected errors.  In the case of instrumentation, we should never cause the app to panic under any circumstances.  So I need to still give some thought if that bit is necessary/warranted at all.

pglombardo:
Merging and going to tag for our first official tagged release of the sensor.

PR #20, opened 2017-04-11
OT spans should be reported using the sdk span type

pglombardo:
Spans are currently reported in generic form.  Using the SDK will allow for better handling, processing and presentation.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -27,11 +27,17 @@ type Span struct {
 }
 
 // NewRecorder Establish a new span recorder
-func NewRecorder(testMode bool) *SpanRecorder {
+func NewRecorder() *SpanRecorder {
 	r := new(SpanRecorder)
-	r.testMode = testMode
 	r.init()
+	return r
+}
 
+// NewTestRecorder Establish a new span recorder used for testing
+func NewTestRecorder() *SpanRecorder {
+	r := new(SpanRecorder)
+	r.testMode = true
+	r.init()
 	return r
 }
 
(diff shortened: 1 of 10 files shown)


pglombardo:
Also, this PR expands on the test suite a bit.  Not complete but a bigger push on end to end tests soon.
screen shot 2017-04-12 at 20 27 04

recorder.go
```diff
@@ -74,24 +96,35 @@ func getHostName(rawSpan basictracer.RawSpan) string {
...
-	s := getStringTag(rawSpan, string(ext.Component))
-	if s == "" {
-		s = getStringTag(rawSpan, string(ext.PeerService))
-		if s == "" {
-			return sensor.serviceName
+	// ServiceName can be determined from multiple sources and has
+	// the following priority (preferred first):
+	//   1. If added to the span via the OT component tag
+	//   2. If added to the span via the OT http.url tag
+	//   3. Specified in the tracer instantiation via Service option
```
boyand: Wouldn't it make sense to change the priority here and have explicitly set service name to be priority 1? The motivation is that component and http.url have different semantical meaning in [OT](https://github.com/opentracing/specification/blob/master/semantic_conventions.md). Also the same service might be producing a number of spans from different components and targeting different urls.

CodingFabian: surprisingly OT has no service naming yet. and for this reason, I would like to get the sdk migration in so that we can make the adaption of OT conventions to instana semantics identical across languages afterwards. In the end, an instana service name is derived on the backend based on properties of a reported span and not by the tracer.

pglombardo: @boyand It's a good point.  Should the tracer setting override the span tags or vice versa?  I'm not sure but it probably doesn't matter for now until the new service mapping support is out.  I'll add a note and revisit.

Issue #22, opened 2017-04-15
Handle case sensitive headers for trace propagation


boyand:
Currently the instana headers are injected lowercase by the carrier implementation. We see problems with this when we deal with child spans between services. The default HTTP implementation in Go would convert any first letter to uppercase https://github.com/golang/go/blob/master/src/net/textproto/reader.go#L554.

Therefore as we read headers from the wire and move them within a context through the service, we may end up with a mixture of uppercase and lowercase headers before the request to the next service is sent. This may often lead to loss of the child span.

Example:

            WIRE      |        READ        |      INJECT
       x-instana-t 1  |  X-Instana-T 1      |   X-Instana-T 1
       x-instana-s 1  |  X-Instana-S 1      |   X-Instana-S 1
                                                x-instana-t 1
                                                x-instana-s 2

pglombardo:
Merged #23.  Thanks @boyand

PR #23, opened 2017-04-15
Handle case-sensitive pre-existing headers

boyand:
This is a quick fix for #22. It introduces a reader to check if headers are already set and reuses the keys. Should work for the currently supported carriers.


pglombardo:
Excellent find & fix.  Thanks @boyand!

PR #25, opened 2017-04-25
first simple event API

pavlobaron:
Bypasses events to the locally running agent using  https://instana.atlassian.net/wiki/display/DOCS/Event+SDK+REST+Web+Service

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,0 +1,47 @@
+package instana
+
+type EventData struct {
+	Title    string `json:"title"`
+	Text     string `json:"text"`
+	Duration int    `json:"duration"`
+	Severity int    `json:"severity"`
+	Plugin   string `json:"plugin,omitempty"`
+	ID       string `json:"id,omitempty"`
+	Host     string `json:"host,omitempty"`
+}
+
+const (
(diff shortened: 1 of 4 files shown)


event.go
```diff
@@ -0,0 +1,47 @@
...
+	ServicePlugin = "com.instana.forge.connection.http.logical.LogicalWebApp"
+	ServiceHost   = ""
+)
+
+func SendDefaultServiceEvent(title string, text string, severity int) {
+	SendServiceEvent(sensor.serviceName, title, text, severity)
+}
+
+func SendServiceEvent(service string, title string, text string, severity int) {
+	sendEvent(&EventData{
```
boyand: Do we need to care about agent state here? I.e. agent.sensor.canSend() or this API is agnostic

pglombardo: This is agnostic.  Can send events at anytime regardless if process is announced or not.  We may want to test error handling eventually though (in the case of no host agent) and make sure things are handled graciously.

pavlobaron: the frequency with which events arrive is not expected to be as high as traces or metrics, plus events API on the agent is generic and doesn't need any dances around pid in different scenarios.

pglombardo:
Tested locally - looks great 👍  thanks @pavlobaron

PR #26, opened 2017-04-25
Fix rsp body closing on error and empty buffer flush to agent

boyand:
* Fixes closing the response body on error too
* Fixes sending empty requests to the agent every second when the buffer is empty
* Moves the agent/recorder state logic to the senders

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -145,18 +145,21 @@ func collectLogs(rawSpan basictracer.RawSpan) map[uint64]map[string]interface{} 
 func (r *SpanRecorder) init() {
 	r.reset()
 
-	if !r.testMode {
-		ticker := time.NewTicker(1 * time.Second)
-		go func() {
-			for range ticker.C {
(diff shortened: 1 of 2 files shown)


recorder.go
```diff
@@ -217,23 +220,25 @@ func (r *SpanRecorder) RecordSpan(rawSpan basictracer.RawSpan) {
 		From:      sensor.agent.from,
 		Data:      &data})
 
-	if !r.testMode && (len(r.spans) == sensor.options.ForceTransmissionStartingAt) {
+	if r.testMode || !sensor.agent.canSend() {
+		return
+	}
+
+	if len(r.spans) >= sensor.options.ForceTransmissionStartingAt {
```
pglombardo: doh! 👍

pglombardo:
Looks great @boyand.  I'll just test manually on a local dev stack as last verification and then merge.  Thanks for the fixes/improvements

PR #27, opened 2017-05-07
Error handling inconsistent/not working

pglombardo:
@boyand  reports that logged errors aren't properly displayed and the span isn't mark as errored.

> it seems OT errors are not handled currently
> the error data we send is not displayed in “data”
> and the span is also not shown as error

- [x] Review/fix error handling/reporting
- [x] Add tests to validate

### DIFF
--- a/json_span.go
+++ b/json_span.go
@@ -12,6 +12,7 @@ type jsonSpan struct {
 	Duration  uint64    `json:"d"`
 	Name      string    `json:"n"`
 	From      *fromS    `json:"f"`
+	Ec        int       `json:"ec,omitempty"`
 	Data      *jsonData `json:"data"`
 }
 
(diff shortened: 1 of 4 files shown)


pglombardo:
While implementing tests for this, I found there are 3 valid/supported methods to indicate an error.  Each of these are now properly translated for the backend:

```Go
import (
	ot "github.com/opentracing/opentracing-go"
	ext "github.com/opentracing/opentracing-go/ext"
	"github.com/opentracing/opentracing-go/log"
)
```

```Go
err := errors.New("Error: File not found")
span.LogFields(log.Error(err), log.String("function", "MyBadfunction"))
```

```Go
span.LogKV("error", "File not found")
```

```Go
ext.Error.Set(span, true)
```

PR #29, opened 2017-05-19
Error on parsing X-Instana-S and X-Instana-T

pglombardo:
X-Instana-S and X-Instana-T headers may contain values between 0 and math.MaxUint64, but textMapPropagator.extract() uses strconv.ParseInt() to parse these headers instead of strconv.ParseUint().

This error cannot be observed if a golang service receives a request from another golang service, because instana/golang-sensor just generates IDs between 0 and math.MaxInt64 by using Rand.Int63(), so these IDs will be parsed well.

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -65,8 +65,16 @@ func (r *textMapPropagator) inject(spanContext ot.SpanContext, opaqueCarrier int
 		return ot.ErrInvalidCarrier
 	}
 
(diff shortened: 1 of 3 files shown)


pglombardo:
Fixed and added a good amount of tests for ongoing validation.  Going to merge for more testing and will tag a release soon.  Thanks again @PSanetra!

PR #38, opened 2017-08-03
Better error reporting

pglombardo:
This PR updates the error detection and logging code to properly
mark the span as errored when a tag or log is set with the key value
of 'error' (or via opentracing/ext.Error).


boyand:
👍

pglombardo:
screen shot 2017-08-04 at 11 33 15

PR #45, opened 2017-10-09
Improved log output: Less noise; proper log levels.

pglombardo:
This PR improves logging output to be less verbose and to output standard state change messages to be in line with our other languages.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -141,7 +141,7 @@ func (r *agentS) fullRequestResponse(url string, method string, data interface{}
 	}
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pglombardo:
Default log output now looks like this:
screen shot 2017-10-09 at 17 21 09

Then if you set the `INSTANA_DEV` environment variable to anything, you get the more verbose version useful for debugging/investigations:
screen shot 2017-10-09 at 17 28 18

Note the proper log levels.  Prior to the PR, it was a bit wild west like.

PR #50, opened 2018-05-04
reuse http client, go fmt

mier85:
reuse the http client because 

// The Client's Transport typically has internal state (cached TCP
// connections), so Clients should be reused instead of created as
// needed. Clients are safe for concurrent use by multiple goroutines.

( copied from the net/http client description )


pglombardo:
Nice find/improvement.  thanks @mier85!

pglombardo:
Merging with the invalid v1.6 test (unsupported version anyways).  Tests updated in https://github.com/instana/golang-sensor/pull/52

Thanks for the improvement @mier85

PR #54, opened 2018-07-25
Support setting the hostname of the agent via environment variable.

doertedev:


agent.go
```diff
@@ -57,13 +58,18 @@ func (r *agentS) makeURL(prefix string) string {
...
+	port := agentDefaultPort
+	if r.sensor.options.AgentPort > 0 {
+		return r.makeFullURL(host, r.sensor.options.AgentPort, prefix)
+	}
+	if envPort == "" {
+		return r.makeFullURL(host, port, prefix)
+	}
+	port, err := strconv.Atoi(envPort)
+	if err != nil {
+		return r.makeFullURL(host, agentDefaultPort, prefix)
```
pglombardo: Shouldn't this use `port` in this call?

doertedev: Not if Atoi error'd - which in case of an error returns 0.

pglombardo: ah good point ok

Issue #55, opened 2018-09-17
User cannot write their own Recorder


etsangsplk:
It provided a public interface but accepts parameter only visible within instana.

Recorder is usspoed to be user configurable setting:
https://github.com/instana/go-sensor/blob/493edb42228321483dd4d59ade71ca93fa89d66b/tracer.go#L113

Recorder is accepting *spanS which end user cannot use because it is private.
https://github.com/instana/go-sensor/blob/493edb42228321483dd4d59ade71ca93fa89d66b/recorder.go#L13

CodingFabian:
Thanks for your contribution. This certainly can be improved, but I am curious why you would want your own recorder. Can you pleasee give a bit of background on the use case so we can support it better?

pglombardo:
We're not going to support user configurable recorders for the foreseeable future.  If there are any concerns or requests about this, please let us know [here](https://support.instana.com/hc/en-us/requests/new).

PR #56, opened 2018-09-17
expose RawSpan on recorder so user can implement their own recorder

etsangsplk:
* make Recorder use public type as parameter for RecordSpan. Only limit the change to maek Recorder work.


pglombardo:
Closing per https://github.com/instana/go-sensor/issues/55#issuecomment-438243329

PR #61, opened 2019-02-15
make intermediate spans possible

steveww:
If a span is not entry nor exit it must be intermediate not ""

At the backend if the type is "" it default to entry, this messes up the flow map with span that should be intermediate.

### DIFF
--- a/span.go
+++ b/span.go
@@ -234,7 +234,7 @@ func (r *spanS) getSpanKind() string {
 	case string(ext.SpanKindRPCClientEnum), "producer", "exit":
 		return "exit"
 	}
-	return ""
+	return "intermediate"
 }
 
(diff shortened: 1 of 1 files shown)


pglombardo:
Thanks Steve - this is helpful.  

I gave you write permissions on this repo.  Going forward, can you clone this repo and work on branches & PRs directly in this repo?  That way I can add in additional fixes when needed.

It seems some of the backend behaviour has evolved so the change you suggest is needed but also a few more - specifically sending the `k` key as well (similar to Python).

I'm going to open another PR and CC you so I can add in my changes as well.

pglombardo:
Closing - fixed in #62.  Thanks Steve!

PR #62, opened 2019-02-15
Span Kind cleanup and improvements

pglombardo:
Better handling & reporting of the `type` and `k` fields in the reported spans.

This replaces #61


pglombardo:
A Trace with an intermediate span:
screen shot 2019-02-15 at 14 43 34

Going to add/update an example that illustrates how to create intermediate spans as well.

pglombardo:
And updated the ot-simple example app with an intermediate span
screen shot 2019-02-15 at 15 08 49

PR #68, opened 2019-04-24
adapters: remove never taken branch

mmlb:
client.Do always returns an error so the type assertion check is unnecessary. See https://play.golang.org/p/GLYZZZXTxux


pglombardo:
Thanks again @mmlb !

mmlb:
:+1:

PR #70, opened 2019-06-16
Add ParentSpanID header propagation

jw-s:
For certain systems like Istio, Parent Span ID propagation is required, hence this PR.


CodingFabian:
Hi Joel,
Would you be so kind to point out where in istio this is necessary? This is not part of B3, Jaeger (or other OT tracer), TraceContext or our Implementation. So aI am curious what istio did there and why they need it.

jw-s:
Hi @CodingFabian, apologies I should have explained further. Istio supports custom mixer adapters which we have to send spans to the instana agent. In the istio instance configuration we need to map the `parentSpanID` field in the tracespan` template to this header as currently you don't propagate this over the wire.

jw-s:
tracespan template: https://istio.io/docs/reference/config/policy-and-telemetry/templates/tracespan/

CodingFabian:
Thanks for the explanation.
It should work without this header just fine. In fact we would silently ignore it.
On the reciving side we will create a new span, join the trace defined in `x-instana-t` and set as parent `x-instana-s` 
But I learned that B3 actually has this header, I did not know that. I think its also unused in b3 ;)

jw-s:
Say I have service A and service B, where service A -> service B. service B will create a new span from service A request, set parentID from service A span but this relationship is only stored inter-process? Since Istio will be dealing with reporting the spans to instana and no longer via the services this relationship will be lost?

CodingFabian:
Service A will create a span, send it to Instana and send `t` and `s` headers through istio.
Now you will create a span in istio, set `t` and parent to the incoming `s`. You send this span to Instana.
For the outgoing call to Service B you set `t` to the same value and send the spanId of your created span as `s` header to B.
B will read the incoming headers and set its parent accordingly and send the span to Instana

you will end up with a trace like this

* `t = 1` , `s=2`, `p=null` from Service A
* `t = 1` , `s=3`, `p=2` from Istio
* `t = 1` , `s=4`, `p=3` from Service B

This is also how our nginx tracing integration works.

jw-s:
Ofcourse, makes sense. I guess this header wouldn't harm but I will close this PR if there's no need as can get this information like you described.

CodingFabian:
There is still a chance I do not fully understand what is needed to build the isio mixer. Please feel free to share with us what you have come up with and if this extension should be needed, we can of course add it!

PR #75, opened 2019-08-08
Added Go module definitions

noctarius:
Adds the Go Module definitions with current versions of the dependencies. Go Module definitions are around since Go 1.11 and are ignored on older Go versions.


pglombardo:
Thanks @noctarius!

PR #76, opened 2019-08-08
Added meaningful additional functions to gain access to the internally…

noctarius:
… (inside the sensor) created tracer, which is necessary to initialize GRPC OT tracing, as well as an additional constructor method, to pass a more complex set of options to create a Sensor

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -13,13 +13,14 @@ import (
 
 type SpanSensitiveFunc func(span ot.Span)
 type ContextSensitiveFunc func(span ot.Span, ctx context.Context)
+type TracerSensitiveFunc func(tracer ot.Tracer)
 
 type Sensor struct {
 	tracer ot.Tracer
 }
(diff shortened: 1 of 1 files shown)


noctarius:
Fixed the missing return, whatever happened there oO

noctarius:
*ping :)

andrewslotin:
Hey @noctarius! Thanks for your contribution! I'm currently working on cleaning up the public interface of this sensor, so I'd rather avoid exposing new methods. 

I totally see the issue with accessing and configuring the underlying tracer. I've just merged https://github.com/instana/go-sensor/pull/86 that allows the caller to inject their own `ot.Tracer` into `instana.Sensor` and to retrieve it later from the parent span as `(opentracing.Span).Tracer()`. This should cover the use cases I can think of, so I will close this PR for now. Please let me know if I've missed something.

PR #78, opened 2019-09-11
added Tracer exposure to adapters

LutzLange:
I need the Tracer to better work with OpenTracing and to be able to connect the go-sensor to OpenTracing GRPC tracing. Please merge.


steveww:
Alternatively set the Opentracing Global Tracer in NewSensor("foo").

InitGlobalTracer(s.tracer)

steveww:
The context contains the parent span that has a Tracer() function to get the tracer.
Not sure this PR is required.
Admittedly the docs could be better.

steveww:
Inside a wrapped handler.

parent := ctx.Value("parentSpan").(ot.Span)
tracer := parent.Tracer()

pglombardo:
@LutzLange does this resolve your issue?  I'll close this PR.

PR #81, opened 2020-02-05
Allow to configure service name via an env variable

### LINKED ISSUE #71: Support INSTANA_SERVICE_NAME environment variable

andrewslotin:
Current implementation of `instana.InitSensor()` immediately falls back to the name of the binary if there was service name provided explicitly via `instana.Options`. These changes allow to configure the service name via the `INSTANA_SERVICE_NAME` env variable, so the name lookup works as follows:

1. `INSTANA_SERVICE_NAME` env var
2. `(instana.Options).Service` value
3. Name of the binary

Closes #71

### DIFF
--- a/sensor.go
+++ b/sensor.go
@@ -61,16 +61,19 @@ func (r *sensorS) configureServiceName() {
 // InitSensor intializes the sensor (without tracing) to begin collecting
 // and reporting metrics.
 func InitSensor(options *Options) {
-	if sensor == nil {
-		sensor = new(sensorS)
-		// If this environment variable is set, then override log level
-		_, ok := os.LookupEnv("INSTANA_DEBUG")
-		if ok {
-			options.LogLevel = Debug
-		}
-
-		sensor.initLog()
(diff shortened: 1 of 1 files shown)


sensor.go
```diff
@@ -44,33 +44,36 @@ func (r *sensorS) setOptions(options *Options) {
...
-
 func (r *sensorS) configureServiceName() {
 	if r.options != nil {
 		r.serviceName = r.options.Service
 	}
 
+	if r.serviceName == "" {
+		r.serviceName = os.Getenv("INSTANA_SERVICE_NAME")
+	}
+
```
pglombardo: One change: The env var should override all.  then `options.Service` and the fallback.

andrewslotin: Currently it already overrides the fallback and my intention was that if something is configured explicitly in the code, it should take the precedence. However if we consider a use case when we need to rename a service without having to change the code, then an env var value prioritized over the option one makes total sense to me. Will adjust 👍

PR #83, opened 2020-02-15
Add automatic profiler

dmelikyan:
Add automatic profiling functionality with examples and docs


noctarius:
Wooooohohohohoooooo 👍 
I feel like I have something to test / hack with next week! :)

autoprofile/auto_profiler.go
```diff
@@ -34,7 +36,14 @@ type AutoProfiler struct {
 	GetExternalPID func() string
 }
 
-func NewAutoProfiler() *AutoProfiler {
+func GetProfiler() *AutoProfiler {
```
andrewslotin: Since the use of `Get` is generally discouraged in Go, can we name it just `Profiler()` or maybe even `DefaultProfiler`?

andrewslotin:
@dmelikyan, I've added some commits on top of your branch, here is a summary of what's been done:
1. I've removed references to `autoprofile.AutoProfiler` instance from samplers. This reference was used mostly to retrieve settings, which are now a part of the samplers types
2. This allowed me to get rid of `autoprofile.AutoProfiler` completely. There are few package-level methods now to configure, enable and disable the profiler.
3. Most of the implementation has been moved to `internal/` until we finalize the public interface of `autoprofiler`.
4. Some cosmetic changes, e.g. more idiomatic error handling, consistent naming, etc.

Let's do one more round of tests and ship it!

autoprofile/internal/sampler_scheduler.go
```diff
@@ -174,7 +174,7 @@ func (ss *SamplerScheduler) startProfiling() bool {
 	}
 	ss.samplerStart = time.Now().UnixNano()
 	ss.samplerTimeout = NewTimer(time.Duration(ss.config.MaxSpanDuration)*time.Second, 0, func() {
-		ss.Stop()
+		ss.stopSampler()
```
andrewslotin: With this we're not going to clear the `started` flag and stop the reporting. Is this intended?

autoprofile/internal/frame_filter.go
```diff
@@ -8,33 +8,20 @@ import (
 )
 
 var (
-	IncludeSensorFrames = true
-	sensorPath          = filepath.Join("github.com", "instana", "go-sensor")
+	IncludeProfilerFrames = false
+	autoprofilePath       = filepath.Join("instana", "go-sensor", "autoprofile")
```
andrewslotin: Did I read this correctly, with this we want to exclude only the profiler frames, but not the whole sensor?

andrewslotin: Also you might consider making this path more strict and keep the `github.com` prefix there just in case this library is used in the code that has `instana/go-sensor/autoprofile` path for some reason.

dmelikyan: Yes, we will exclude only profiler stacks for now. We do not automatically instrument other parts of sensor, so there is not much to exclude.
Regarding the path being more strict, good point, I'll add the github.com back, it was removed for local testing.

autoprofile/internal/sampler_scheduler.go
dmelikyan: `Start` and `Stop` are used to start and stop the scheduler, not the sampling itself.

andrewslotin: Right, I see 👍

PR #86, opened 2020-03-09
Span propagation in `context.Context`

andrewslotin:
This PR introduces an API to propagate the active span via `context.Context` to standardize the way of storing and retrieving the parent span throughout different instrumentations.


andrewslotin:
@pglombardo, good question! I'm not sure if this API is worth to be mentioned in the README. The main goal for it is to provide a common way for the upcoming integrations to store and retrieve the active span in `context.Context`.

noctarius:
I think a short paragraph on how the context transfers the current span may be interesting in the sense of others being able to provide pull requests with additional wrappers :-)

andrewslotin:
@pglombardo added a new section on trace context propagation.

PR #89, opened 2020-03-13
Add instrumentation for google.golang.org/grpc library

andrewslotin:
This PR adds a new module `github.com/instana/go-sensor/instrumentation/instagrpc` that provides an instrumentation for GRPC servers and clients that use `google.golang.org/grpc`.


andrewslotin:
@pglombardo, good point. I've added some tests, but still worth to do check it against the actual installation 👍

PR #91, opened 2020-03-19
Instrument net/http

andrewslotin:
This is a new approach to instrumenting HTTP handlers and clients using `instana.RoundTripper()` and `instana.TracingHandlerFunc()`:

* `instana.TracingHandlerFunc()` is used to instrument HTTP handlers. It's implementation is almost identical to `(*instana.Sensor).TracingHandler()`, but takes a sensor instance as a parameter instead of a receiver for consistency with `instrumentation/instagrpc`
* `instana.RoundTripper()` allows to instrument the standard `http.Client` once at the top level and does not require an instance of `instana.Sensor` to make requests.

This implementation also addresses HTTP error messages not being shown in the "Error Messages" tab and call details pane.

This PR deprecates the use of `instana.Sensor` methods for instrumentation.


instrumentation.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		span := tracer.StartSpan(name, opts...)
+		defer span.Finish()
+
+		defer func() {
+			// Be sure to capture any kind of panic / error
+			if err := recover(); err != nil {
+				if e, ok := err.(error); ok {
+					span.SetTag("message", e.Error())
```
pglombardo: Do we really need the duplication?  The span has http tags so it will always be converted to an HTTPSpan in the backend - so only `http.error` needed right?

andrewslotin: I did a manual test and there seem be the same issue as we had with `rpc.error`: the error message from `http.error` is not being used as `message` and the error calls show up as an error without message.

andrewslotin: Here are examples of an erroneous call made with `http.error` only and with both `http.error` and `message` fields populated:
Screenshot 2020-03-24 at 13 48 33

Issue #13, opened 2017-03-16
Needed: A better logging strategy


pglombardo:
Pointed out by @monolovl in #2 the logging could be improved in this package.

To be consistent with our other languages we need a logger that:

1. Supports Debug/Warn/Info/Fatal levels
2. Configurable in-code & replaceable by app logger
3. Configurable via environment variables
4. Defaults to stdout/stderr but configurable
5. Messages prefixed with `[instana]` or equivalent

General goal for instrumentation is to:

1. Boot a single boot message and not print anything else if all goes well
2. In case of error, print a _single_ error and handle in best case
3. If handling is not possible, deactivate the sensor and sit in a corner

mier85:
the instana prefix shouldn't be hardcoded. We use structured logging, so we would add a field to the logger that says "lib":"instana" or something like that instead of having it inside the log message itself

PR #93, opened 2020-03-24
Pluggable logger

andrewslotin:
This PR introduces the new `github.com/instana/go-sensor/logger` package that provides a leveled logger implementation used by Go sensor and autoprofiler to write diagnostic and error messages. This is **NOT** logging instrumentation.

An instance of `logger.Logger` is used as a default, has configurable prefix and respects `INSTANA_DEBUG` environment variable, overriding any log level set in the code with `logger.DebugLevel`.

I've also added two new methods: `instana.SetLogger()` and `autoprofile.SetLogger()` that allow to use any 3rd party logger that conforms the `instana.LeveledLogger` interface, e.g. `github.com/sirupsen/logrus.Logger` or `go.uber.org/zap.SugaredLogger`. In this case the `INSTANA_DEBUG` variable needs to be handled explicitly, as different loggers may have different ways to set the min logging level.

This PR deprecates `autoprofile.SetLogLevel()` in favor of `autoprofile.SetLogger()`.

Closes https://github.com/instana/go-sensor/issues/13

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,63 @@ var sensor = instana.NewSensor("my-service")
 
 A full example can be found under the examples folder in [example/webserver/instana/http.go](./example/webserver/instana/http.go).
 
+### Log Output
+
+The Go sensor uses a leveled logger to log internal errors and diagnostic information. The default `logger.Logger` uses `log.Logger`
+configured with `log.Lstdflags` as a backend and writes messages to `os.Stderr`. By default this logger only prints out the `ERROR` level
+messages unless the environment variable `INSTANA_DEBUG` is set.
+
+To change the min log level in runtime it is recommended to configure and inject an instance of `instana.LeveledLogger` instead of using
+the deprecated `instana.SetLogLevel()` method:
+
+```go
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -24,6 +24,63 @@ var sensor = instana.NewSensor("my-service")
 
 A full example can be found under the examples folder in [example/webserver/instana/http.go](./example/webserver/instana/http.go).
 
+### Log Output
+
+The Go sensor uses a leveled logger to log internal errors and diagnostic information. The default `logger.Logger` uses `log.Logger`
+configured with `log.Lstdflags` as a backend and writes messages to `os.Stderr`. By default this logger only prints out the `ERROR` level
```
mmanciop: @andrewslotin we usually collect `WARN` too

andrewslotin: @mmanciop, good to know! Will keep it in mind when we start working on log capturing 😃 

These changes, however, are irrelevant and only address the way the Go sensor writes internal diagnostic messages. At the moment it uses an internal logger that outputs to stderr in plain text. Once this PR is merged it'll be possible to configure this behavior or even use an external logger.

PR #96, opened 2020-04-02
Add support for registered span format

andrewslotin:
Apart from SDK spans carrying arbitrary data, Instana agent supports registered spans for various types of activities within the app. These spans have stricter format and generally less error-prone since the backend does not have to perform a conversion on its side.

This PR introduces support for registered spans to the Go sensor. Whenever an operation name matches a known `instana.RegisteredSpanType` value, the respective span data type will be used for populating the `(instana.jsonSpan).Data` field.

The following steps need to be taken to add support for an already registered span:
1. Add a new `RegisteredSpanType` constant for this operation type
2. Add a constructor function that extracts necessary data from the tracer tag and returns a struct to be used for serialization.

I've already included support for RPC and HTTP spans in these changes, however for them to take effect in `instagrpc` I'd need to release a new version of `go-sensor` first, so the installations that use Go modules could pick up the new types.


json_span.go
```diff
@@ -1,39 +1,227 @@
...
+		case "rpc.error":
+			readStringTag(&tags.Error, v)
+		}
+	}
+
+	return tags
+}
+
+func readStringTag(dst *string, tag interface{}) {
+	if v, ok := tag.(string); ok {
```
pglombardo, line 218: Can it happen that the tag may not be string (or the expected type)?

andrewslotin: The `dst` won't be populated. The problem is that the OT span tags are not strictly typed, so we have to do our best. However it might make sense to handle `[]byte` here as well and various `int` types in the `readIntTag`. What do you think?

pglombardo: Ok as long as we don't crash/error out.  Missing tags can be fixed if found later...

PR #109, opened 2020-04-23
Add suppress tracing support

andrewslotin:
This PR makes Go sensor respecting the `X-INSTANA-L` header value. The sensor suppresses tracing when this header is set to `"0"` and won't send the span created from this context or any of its child spans to the agent. This value is forwarded downstream to signal that dependent services should also suppress tracing on their side.

To suppress the trace starting from a user-created span, Go sensor provides a new `instana.SuppressTracing()` tag. Once set, it prevents the span and its children from being sent to the agent.

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -8,8 +8,6 @@ import (
 	ot "github.com/opentracing/opentracing-go"
 )
 
-type textMapPropagator struct{}
-
 // Instana header constants
 const (
 	// FieldT Trace ID header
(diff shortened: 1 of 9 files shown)


propagation.go
```diff
@@ -24,12 +19,7 @@ const (
 	FieldB = "x-instana-b-"
 )
 
-func (r *textMapPropagator) inject(spanContext ot.SpanContext, opaqueCarrier interface{}) error {
-	sc, ok := spanContext.(SpanContext)
-	if !ok {
-		return ot.ErrInvalidSpanContext
-	}
-
+func injectTraceContext(sc SpanContext, opaqueCarrier interface{}) error {
```
pglombardo, line 22: The OT spec says that these methods should be named inject/extract for compatibility.  The OTel specification seems to have kept the naming: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#inject

Although I like your naming scheme better, I believe this will break compatibility.

andrewslotin: Not really, we provide `instana.tracerS` to use as a propagator. This type satisfies [`opentracing.Tracer`](https://godoc.org/github.com/opentracing/opentracing-go#Tracer) described in the spec you've mentioned. `{inject,extract}TraceContext()` are just internal methods used by `(instana.tracerS).{Inject,Extract}()`.

PR #128, opened 2020-06-22
Introduce INSTANA_AUTO_PROFILE

mmanciop:
Add option to activate AutoProfile via the `INSTANA_AUTO_PROFILE` env car, overriding the compiled setup. This also should work to enable AutoProfile with just recompilation and adding `INSTANA_AUTO_PROFILE=true` to the environment.

### DIFF
--- a/sensor.go
+++ b/sensor.go
@@ -3,6 +3,7 @@ package instana
 import (
 	"errors"
 	"os"
+	"strings"
 
 	"github.com/instana/go-sensor/autoprofile"
 	"github.com/instana/go-sensor/logger"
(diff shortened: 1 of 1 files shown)


sensor.go
```diff
@@ -82,7 +83,7 @@ func InitSensor(options *Options) {
 	sensor = newSensor(options)
 
 	// enable auto-profiling
-	if options.EnableAutoProfile {
+	if options.EnableAutoProfile || strings.ToLower(os.Getenv("INSTANA_AUTO_PROFILE")) == "true" {
```
andrewslotin, line 86: I think we could simplify configuration and simply check whether `INSTANA_AUTO_PROFILE` is set, just like we do with `INSTANA_DEBUG`.

andrewslotin:
Replaced by https://github.com/instana/go-sensor/pull/132

PR #136, opened 2020-07-14
Add instrumentation for database/sql

andrewslotin:
This PR adds instrumentation for database calls made with `database/sql` package. It introduces two way of instrumenting the DB client, depending on the way it's initialized:

* For the code that establishes the database connection using `sql.Open()` the Go sensor is now offering the `instana.SQLOpen()` convenience wrapper that uses the instrumented version of a database driver registered with `instana.InstrumentSQLDriver()`.
* In case `sql.OpenDB()` (introduced in Go v1.10) is used to initialize a database client, the `driver.Connector` needs to be wrapped with `instana.WrapSQLConnector()` before passing it to the `sql.OpenDB()` call.

The following information is being collected and sent with each database span:
* SQL query
* Database name (schema, collection, etc.)
* Database server host and port
* Connection string used to establish the connection


andrewslotin:
That's a good point, I'll try to extract the instrumentation into a separate package and see if we have any dependencies on non-exported methods and types in `instana`.

andrewslotin:
@pglombardo, I ended up with your suggestion to rename `instana.OpenSQLDB()` to `instana.SQLOpen()` that reads nicer indeed. Extracting the `database/sql` instrumentation code into a separate package is possible, however I found the package name collisions confusing even while updating the tests.

instrumentation_sql.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil && err != driver.ErrSkip {
+		sp.LogFields(otlog.Error(err))
+	}
+
+	return res, err
+}
+
+func startSQLSpan(ctx context.Context, conn dbConnDetails, query string, sensor *Sensor) ot.Span {
+	tags := ot.Tags{
+		string(ext.DBType):      "sql",
```
YoannBuch, line 223: @andrewslotin I would be nice to set the database type whenever it's known. For example if the mysql strategy was used to parse the connection string, we could send "mysql".

andrewslotin: The OpenTracing spec [prescribes](https://opentracing.io/specification/conventions/) the use of `sql` for any SQL database:

> Database type. For any SQL database, "sql". For others, the lower-case database category, e.g. "cassandra", "hbase", or "redis".

YoannBuch: right so "mysql" if we know it's mysql, etc. and then "sql" otherwise.

andrewslotin: Not exactly. If this is an SQL database, then "sql". Otherwise put the lower-case database category, e.g. "mongo" or "cassandra".

YoannBuch: sorry read too fast, again... thanks for being patient ;)

PR #138, opened 2020-07-20
Support EUM script correlation

andrewslotin:
Forward the correlation ID and type provided by Instana end-user monitoring script to the host agent in `span.crid` and `span.crtp` fields. These values override provided `X-Instana-T` and `X-Instana-S` header values, effectively forcing a new trace when provided.

### DIFF
--- a/tracer_test.go
+++ b/tracer_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	instana "github.com/instana/go-sensor"
+	ot "github.com/opentracing/opentracing-go"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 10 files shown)


tracer_test.go
```diff
@@ -42,3 +43,18 @@ func TestTracer_StartSpan_SuppressTracing(t *testing.T) {
 	sc := sp.Context().(instana.SpanContext)
 	assert.True(t, sc.Suppressed)
 }
+
+func TestTracer_StartSpan_WithCorrelationData(t *testing.T) {
+	recorder := instana.NewTestRecorder()
+	tracer := instana.NewTracerWithEverything(&instana.Options{}, recorder)
+
+	sp := tracer.StartSpan("test", ot.ChildOf(instana.SpanContext{
```
pglombardo, line 51: Is this a valid use case? I didn't know that...

andrewslotin: `ot.ChildOf()` takes a span context, which is an interface type, so there it's possible to use any type that satisfies `ot.SpanContext` there, although you probably should not create your own parent contexts in production code... Here is just for testing purposes only.

PR #141, opened 2020-08-12
Collect HTTP request parameters

andrewslotin:
This PR updates the HTTP server and client instrumentation to collect and send the query string parameters to the agent. To avoid leaking sensitive data, the tracer now supports secrets matcher provided in `(instana.Options).Tracer.Secrets`. This matcher can be specified during the sensor initialization or in [the Host Agent Configuration file](https://www.instana.com/docs/setup_and_manage/host_agent/configuration/#secrets) and filters out any HTTP parameters which name match the criteria. The host agent configuration takes precedence.

The default filtering rule is `contains-ignore-case` with the following list of match terms: `key`, `password` and `secret`. If an HTTP parameter name contains any of these strings regardless of case, it's value will be replaced with `<redacted>` before sent to the host agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ The Instana Go sensor consists of two parts:
 * [Common Operations](#common-operations)
   * [Setting the sensor log output](#setting-the-sensor-log-output)
   * [Trace Context Propagation](#trace-context-propagation)
+  * [Secrets Filtering](#secrets-filtering)
   * [HTTP servers and clients](#http-servers-and-clients)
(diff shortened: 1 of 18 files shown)


andrewslotin:
@pglombardo, good point. The proposed behavior is consistent with our Node and PHP sensors, but I see that for Python it's different and you actually have a good point that we should be filtering the value, not the whole parameter.

basti1302:
I agree with Peter, only redacting the value is the superior user experience (even if that's not how it is implemented in Node currently)

PR #215, opened 2021-04-06
Correct typos in readme

andriikushch:
There are a few minor typos and URLs in the example section fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ http.HandleFunc("/", instana.TracingHandlerFunc(sensor, "/", func(w http.Respons
 }))
 ```
 
-In case your handler is implemented as a http.Handler, pass its `ServeHTTP` method instead:
+In case your handler is implemented as a `http.Handler`, pass its `ServeHTTP` method instead:
 
 ```go
 h := http.FileServer(http.Dir("./"))
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Services running in serverless environments don't use host agent to send metrics
 | `INSTANA_TAGS`               | none                                       | A comma-separated list of tags with optional values to associate with the ECS task       |
 | `INSTANA_ZONE`               | `<Current AWS availability zone>`          | A custom Instana zone name for this service                                              |
 
-Please refer to [Insana documentation](https://www.instana.com/docs/reference/environment_variables/#serverless-monitoring) for more detailed description of these variables and their value format.
+Please refer to [Instana documentation](https://www.instana.com/docs/reference/environment_variables/#serverless-monitoring) for more detailed description of these variables and their value format.
 
 ### Using Instana to gather process metrics only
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,9 +200,9 @@ func MyFunc(ctx context.Context) {
...
+Certain instrumentation provided by the Go sensor package, e.g. the [HTTP servers and clients](#http-servers-and-clients) wrappers, collect data that may contain sensitive information, such as passwords, keys and secrets. To avoid leaking these values the Go sensor replaces them with `<redacted>` before sending to the agent. The list of parameter name matchers is defined in `com.instana.secrets` section of the [Host Agent Configuration file](https://www.instana.com/docs/setup_and_manage/host_agent/configuration/#secrets) and will be sent to the in-app tracer during the announcement phase (requires agent Go trace plugin `com.instana.sensor-golang-trace` v1.3.0 and above).
```
andrewslotin, line 203:
```suggestion
Certain instrumentation modules provided by the Go sensor package, e.g. the [HTTP servers and clients](#http-servers-and-clients) wrappers, collect data that may contain sensitive information, such as passwords, keys and secrets. To avoid leaking these values the Go sensor replaces them with `<redacted>` before sending to the agent. The list of parameter name matchers is defined in `com.instana.secrets` section of the [Host Agent Configuration file](https://www.instana.com/docs/setup_and_manage/host_agent/configuration/#secrets) and will be sent to the in-app tracer during the announcement phase (requires agent Go trace plugin `com.instana.sensor-golang-trace` v1.3.0 and above).
```
I believe this would be more accurate term

README.md
```diff
@@ -200,9 +200,9 @@ func MyFunc(ctx context.Context) {
...
 
-The default setting for the secrets matcher is `contains-ignore-case` with following list of terms: `key`, `password`, `secret`. This would redact the value of a parameter which name _contains_ any of these strings ignoring the case.
+The default setting for the secret's matcher is `contains-ignore-case` with following list of terms: `key`, `password`, `secret`. This would redact the value of a parameter which name _contains_ any of these strings ignoring the case.
```
andrewslotin, line 205: Here the use of plural "secrets" is intentional, as this matcher is used to match all secrets, not a specific one.

README.md
```diff
@@ -234,7 +234,7 @@ http.HandleFunc("/", instana.TracingHandlerFunc(sensor, "/", func(w http.Respons
 }))
 ```
 
-In case your handler is implemented as an http.Handler, pass its `ServeHTTP` method instead:
+In case your handler is implemented as a http.Handler, pass its `ServeHTTP` method instead:
```
andrewslotin, line 237:
```suggestion
In case your handler is implemented as a `http.Handler`, pass its `ServeHTTP` method instead:
```
Let's apply some formatting since we're on it :)

andrewslotin (changes requested):
Well spotted 👍 I made a couple of suggestions to use make terminology consistent with our docs and other tracers.

README.md
andriikushch: Agree 👍

README.md
andriikushch: Fine for me. I will revert it.

README.md
andriikushch: 👍

andrewslotin (approved):
Excellent! 👍

PR #217, opened 2021-04-07
Update READMEs and method receiver names

andriikushch:
There are a few minor typos fixed. And some method receivers are renamed because they had different names in different places.

### DIFF
--- a/instrumentation/cloud.google.com/go/pubsub/README.md
+++ b/instrumentation/cloud.google.com/go/pubsub/README.md
@@ -17,11 +17,11 @@ $ go get github.com/instana/go-sensor/instrumentation/cloud.google.com/go/pubsub
 Usage
 -----
 
-This module is a drop-in replacement for `cloud.google.com/go/pubsub` in common publisher/subscriber use cases. It provides
+This module is a drop-in replacement for `cloud.google.com/go/pubsub` in a common publisher/subscriber use cases. It provides
 type aliases for `pubsub.Message`, `pubsub.PublishResult` and `pubsub.Subscription`, however if your code performs any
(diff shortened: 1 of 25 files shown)


instrumentation/cloud.google.com/go/pubsub/README.md
```diff
@@ -17,11 +17,11 @@ $ go get github.com/instana/go-sensor/instrumentation/cloud.google.com/go/pubsub
 Usage
 -----
 
-This module is a drop-in replacement for `cloud.google.com/go/pubsub` in common publisher/subscriber use cases. It provides
+This module is a drop-in replacement for `cloud.google.com/go/pubsub` in a common publisher/subscriber use cases. It provides
```
andrewslotin, line 20: Not sure "a" is needed, since we're talking about plural numbers.

instrumentation/instasarama/partition_consumer_test.go
```diff
@@ -54,12 +54,12 @@ func TestPartitionConsumer_Messages(t *testing.T) {
 	var collected []*sarama.ConsumerMessage
 	timeout := time.After(1 * time.Second)
 
-CONSUMER_LOOP:
+ConsumerLoop:
```
andrewslotin, line 57: I'd rather suggest we keep the label the way it is to not to confuse with a variable/constant.

andrewslotin (changes requested):
I'd suggest we keep the label upper case

instrumentation/cloud.google.com/go/pubsub/README.md
andriikushch: Ok, I thought it belongs to the publisher. But I got your point :) I'll revert it

instrumentation/instasarama/partition_consumer_test.go
andriikushch: Fine for me. I'll revert

andrewslotin (approved):
👍

PR #218, opened 2021-04-08
Add example of grpc server and client

andriikushch:


example/grpc-example/main.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+const (
+	defaultPort    = ":43210"
+	defaultAddress = "localhost"
+	testMessage    = "Hi!"
+)
+
+func main() {
+	// `port` will be used by an example
```
andrewslotin, line 30: Let's omit the comments that are not related to instrumentation with Instana

example/grpc-example/main.go
```diff
@@ -0,0 +1,149 @@
...
+// Service is used to implement pb.EchoService
+type Service struct {
+	pb.UnimplementedEchoServiceServer
+}
+
+// Echo implements pb.EchoService
+func (s *Service) Echo(ctx context.Context, in *pb.EchoRequest) (*pb.EchoReply, error) {
+	log.Printf("Request << %s", in.GetMessage())
+
+	// Extract the parent span and use its tracer to initialize any child spans to trace the calls
```
andrewslotin, line 129: Let's emphasize that this is purely optional and not required for minimal instrumentation

andrewslotin (changes requested):
Overall looks good! I think we could simplify this example even more and make sure that we only keep comments that are specific to instrumentation. This is code is to demonstrate the use of Instana after all, and not how to implement a gRPC server 😃

example/grpc-example/main.go
andriikushch: Makes sense, I will change it.

example/grpc-example/main.go
andriikushch: This is done.

example/grpc-example/main.go
andriikushch: This is done.

andrewslotin (approved):
👍

PR #219, opened 2021-04-15
Extended w3 c trace context support in aws lambda

andriikushch:
This PR enables the use of w3 trace context headers, in case of Instana headers are absent. 

It is done by increasing the version of the go-sensor package which already contains all necessary logic.

In addition to that, there are few test cases, that test this behavior.


andrewslotin (commented):
Good job! Shall we also add a test to make sure that Instana trace headers are preferred over the W3C ones?

andrewslotin (approved):
👍

PR #220, opened 2021-04-19
When Instana headers were corrupted, use w3Context and do not return en error

andriikushch:
Proposal to not return an error (ot.ErrSpanContextCorrupted) while extracting a trace context, for a case:

- when the context is not suppressed
- Instana headers either can not be parsed or present partially

But **w3 context is there**

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -114,6 +114,9 @@ func injectTraceContext(sc SpanContext, opaqueCarrier interface{}) error {
 	return nil
 }
 
+// This method searches for Instana headers (FieldT, FieldS, FieldL and header with name prefixed with FieldB)
+// and try to parse their values. It also tries to extract w3 context and assign it inside returned object. W3 context
+// will be propagated further and can be used as a fallback.
 func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 	spanContext := SpanContext{
 		Baggage: make(map[string]string),
@@ -128,6 +131,8 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 		pickupW3CTraceContext(http.Header(c), &spanContext)
 	}
 
+	// Iterate over the headers, look for Instana headers and try to parse them.
+	// In case of error interrupt, iteration and return it.
 	err := carrier.ForeachKey(func(k, v string) error {
 		var err error
 
@@ -135,14 +140,18 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 		case FieldT:
 			spanContext.TraceIDHi, spanContext.TraceID, err = ParseLongID(v)
 			if err != nil {
+				sensor.logger.Debug("extract trace context ", FieldT, ": ", err)
 				return ot.ErrSpanContextCorrupted
 			}
 		case FieldS:
 			spanContext.SpanID, err = ParseID(v)
 			if err != nil {
+				sensor.logger.Debug("extract trace context ", FieldS, ": ", err)
 				return ot.ErrSpanContextCorrupted
 			}
 		case FieldL:
+			// When FieldL is presented and equal to "0", then spanContext is suppressed.
+			// In addition to that non-empty correlation data may be extracted.
 			suppressed, corrData, err := parseLevel(v)
 			if err != nil {
 				sensor.logger.Info("failed to parse ", k, ": ", err, " (", v, ")")
@@ -167,9 +176,9 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 		return spanContext, err
 	}
 
-	// reset the trace if a correlation ID has been provided
+	// reset the trace IDs if a correlation ID has been provided
 	if spanContext.Correlation.ID != "" {
-		spanContext.TraceIDHi, spanContext.TraceID, spanContext.SpanID = 0, 0, 0
+		resetSpanContextIDs(&spanContext)
 
 		return spanContext, nil
 	}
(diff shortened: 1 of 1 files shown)


propagation.go
```diff
@@ -178,7 +178,13 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 		return spanContext, ot.ErrSpanContextNotFound
 	}
 
-	if !spanContext.Suppressed && (spanContext.SpanID == 0 != (spanContext.TraceIDHi == 0 && spanContext.TraceID == 0)) {
+	// when the context is not suppressed
+	// and instana headers either can not be parsed or present partially
+	// and w3 context is not there
+	if !spanContext.Suppressed &&
+		(spanContext.SpanID == 0 != (spanContext.TraceIDHi == 0 && spanContext.TraceID == 0)) &&
+		spanContext.W3CContext.IsZero() {
```
andrewslotin, line 186: So if there is a non-suppressed request that has a W3C trace and a _broken_ Instana context, we'll return a span context with inconsistent Instana IDs and a valid W3C trace context, right? Since Instana IDs always take precedence when present, we'll use these broken values to continue the trace and probably break it.

I believe in this case we should give up using Instana trace context, probably log this error and continue as if there are no `X-Instana-` headers have been set.

andrewslotin (changes requested):

propagation.go
andriikushch: Thank you for your feedback, I think your suggestions make sense. I've changed the code.

- Now if only some Instana ID headers present and the w3 context is not zero, I do reset context IDs and continue without error. 
- I also added logging for the mentioned case and in two more places (when parsing headers) to see if there any issue with the header format/value. I am using a debug level to avoid writing a huge amount of logs in case of misconfiguration.

propagation.go
```diff
@@ -128,21 +131,27 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
...
+	// Iterate over the headers, look for Instana headers and try to parse them.
+	// In case of error interrupt, iteration and return it.
 	err := carrier.ForeachKey(func(k, v string) error {
 		var err error
 
 		switch strings.ToLower(k) {
 		case FieldT:
 			spanContext.TraceIDHi, spanContext.TraceID, err = ParseLongID(v)
 			if err != nil {
+				sensor.logger.Debug("extract trace context ", FieldT, ": ", err)
```
andrewslotin, line 143: I guess this debug log is intended for debugging the case when headers are broken? In this case it might make sense to include the original headers into the same message, otherwise there is a high chance these two lines will be impossible find in the log.

propagation.go
```diff
@@ -178,13 +187,30 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
...
+			" TraceID=", spanContext.TraceID,
+			" TraceIDHi=", spanContext.TraceIDHi)
+
+		// Check if w3 context was found
+		if spanContext.W3CContext.IsZero() {
+			return spanContext, ot.ErrSpanContextCorrupted
+		}
+
+		// Reset the trace IDs
+		resetSpanContextIDs(&spanContext)
```
andrewslotin, line 204: I'd rather suggest we explicitly return a new `SpanContext` here that has only the W3C trace and not reuse the existing one. I believe this way it'd be easier to follow the logic and also allow us to not to maintain the `resetSpanContextIDs()` in case we add now fields to the context.

propagation.go
```diff
@@ -178,13 +187,30 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
 		return spanContext, ot.ErrSpanContextNotFound
 	}
 
-	if !spanContext.Suppressed && (spanContext.SpanID == 0 != (spanContext.TraceIDHi == 0 && spanContext.TraceID == 0)) {
-		return spanContext, ot.ErrSpanContextCorrupted
+	// When the context is not suppressed and one of Instana ID headers set.
+	if !spanContext.Suppressed &&
+		(spanContext.SpanID == 0 != (spanContext.TraceIDHi == 0 && spanContext.TraceID == 0)) {
+		sensor.logger.Debug("only one of Instana ID headers present",
```
andrewslotin, line 193: This might be a personal preference, but shall we make it clearer that something is wrong by using words like "broken" or "corrupted"? Something like "broken instana trace context" and format ids there using `instana.Formap{ID,LongID}" since it's more familiar to the user.

andrewslotin (changes requested):

propagation.go
andriikushch: Agree 👍  I add it.

propagation.go
andriikushch: 👍  I will change it

propagation.go
andriikushch: Makes sense what you suggested. The initial idea was to be very specific about the issue, but from another side, it was not clear that something is wrong 😃

propagation.go
```diff
@@ -128,21 +131,27 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
...
+	// Iterate over the headers, look for Instana headers and try to parse them.
+	// In case of error interrupt, iteration and return it.
 	err := carrier.ForeachKey(func(k, v string) error {
 		var err error
 
 		switch strings.ToLower(k) {
 		case FieldT:
 			spanContext.TraceIDHi, spanContext.TraceID, err = ParseLongID(v)
 			if err != nil {
+				sensor.logger.Debug("extract trace context ", FieldT, "=", v, " : ", err)
```
andrewslotin, line 143: Do we still need these log lines?

propagation.go
```diff
@@ -178,8 +187,23 @@ func extractTraceContext(opaqueCarrier interface{}) (SpanContext, error) {
...
+	if !spanContext.Suppressed &&
+		(spanContext.SpanID == 0 != (spanContext.TraceIDHi == 0 && spanContext.TraceID == 0)) {
+		sensor.logger.Debug("broken Instana trace context:",
+			" SpanID=", FormatID(spanContext.SpanID),
+			" TraceID=", FormatLongID(spanContext.TraceIDHi, spanContext.TraceID))
+
+		// Check if w3 context was found or not
+		if spanContext.W3CContext.IsZero() {
+			return spanContext, ot.ErrSpanContextCorrupted
+		} else {
```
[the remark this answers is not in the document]
andrewslotin: Also I'd rather flip the condition and return the `SpanContext{ W3CContext: spanContext.W3CContext }` in case `spanContext.W3CContext` is not zero, and an error by default, just for readability reasons.

propagation.go
```diff
@@ -114,6 +114,9 @@ func injectTraceContext(sc SpanContext, opaqueCarrier interface{}) error {
 	return nil
 }
 
+// This method searches for Instana headers (FieldT, FieldS, FieldL and header with name prefixed with FieldB)
+// and try to parse their values. It also tries to extract w3 context and assign it inside returned object. W3 context
```
andrewslotin, line 118:
```suggestion
// and try to parse their values. It also tries to extract W3C context and assign it inside returned object. W3C context
```
C stands for consortium there 😉

andrewslotin (changes requested):
Just a few minor comments on typos and readability. Otherwise it looks good to me 👍

propagation.go
andriikushch: Without it, there is no way to see the original error from `ParseLongID `. Assuming this is debugging level, I would keep it. Anyway, this error should never exist in a healthy environment.

propagation.go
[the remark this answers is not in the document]
andriikushch: 👍

propagation.go
andriikushch: I thought once more, let's remove it :)

andrewslotin: You’re right, this might be helpful while debugging, however in case of handling concurrent requests this log line may be far away from the one where we report an error and if, for example, the header level is a non-hex string, there is no way to correlate these two messages. This is why I suggested to put the malformed value into the error log at first place. But I guess logging the parsed values is also fine, at least we know which piece of context is missing

andrewslotin (approved):
Perfect!

PR #221, opened 2021-04-19
always read a response body, to enable connection reuse

andriikushch:
Reading a Body together with closing it will allow the reuse of underlying network resources. According to https://github.com/golang/go/blob/a72622d028077643169dc48c90271a82021f0534/src/net/http/client.go#L562

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -280,7 +280,7 @@ func (agent *agentS) fullRequestResponse(url string, method string, data interfa
 					}
 				}
 
-				io.Copy(io.Discard, resp.Body)
+				io.Copy(ioutil.Discard, resp.Body)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -278,6 +279,8 @@ func (agent *agentS) fullRequestResponse(url string, method string, data interfa
 						ret = resp.Header.Get(header)
 					}
 				}
+
+				io.Copy(ioutil.Discard, resp.Body)
```
andrewslotin, line 283: Makes sense, let's put a limit here though, to avoid reading unreasonably large responses though:
```suggestion
				io.CopyN(ioutil.Discard, resp.Body, 256 << 10)
```
The constant comes from [here](https://golang.org/src/net/http/server.go#L1097) and corresponds an average TCP buffer size.

andrewslotin (changes requested):

agent.go
andriikushch: Good point!

andrewslotin (approved):

PR #222, opened 2021-04-19
Add custom tags support to instrumentation spans

andrewslotin:
This PR updates the Go tracer to send user-defined tags to the tags created by instrumentation packages, for example the ones created for HTTP client/server requests. These tags are displayed in the call details and can be used to search for calls in the [Unbound Analytics](https://www.instana.com/docs/unbounded_analytics/#main) view.


registered_span.go
```diff
@@ -0,0 +1,1056 @@
...
+
+// NewHTTPSpanData initializes a new HTTP span data from tracer span
+func NewHTTPSpanData(span *spanS) HTTPSpanData {
+	data := HTTPSpanData{
+		SpanData: NewSpanData(span, RegisteredSpanType(span.Operation)),
+		Tags:     NewHTTPSpanTags(span),
+	}
+
+	kindTag := span.Tags[string(ext.SpanKind)]
+	data.clientSpan = kindTag == ext.SpanKindRPCClientEnum || kindTag == (ext.SpanKindRPCClientEnum)
```
andriikushch, line 200: It seems there is a typo

```suggestion
	data.clientSpan = kindTag == ext.SpanKindRPCClientEnum || kindTag == string(ext.SpanKindRPCClientEnum)
```

andriikushch (changes requested):
Overall looks good to me 👍 . Please check my small suggestion and I am ready to approve.

registered_span.go
andrewslotin: Well spotted, I believe there should be no type cast though

andrewslotin: My bad, I misread this change :) You're right!

andriikushch (approved):
Looks good to me 😃  Let's merge it 🚀

PR #223, opened 2021-04-22
Requeue spans safely

andrewslotin:
This PR updates the span collector to safely re-queue spans in case the submission attempt has failed by acquiring the lock before modifying the span queue and putting them to the beginning of the queue, so they are evicted first should we truncate the buffer.


andriikushch (approved):
Looks good to me 👍

PR #225, opened 2021-04-25
(WIP) pgxpool support

noctarius:
This contribution adds a support wrapper for pgxpool. Two different execution models are supported. Pool is a direct wrapper of the underlying pgxpool Pool instance and just captures the query time, whereas everything in extensions.go offers utility methods which use the instrumented Pool instance to capture query and fetch time (where query+fetch time is implemented as an internal span in Instana).


sanojsubran:
Closing this draft PR as it has been dormant for quite sometime.

PR #227, opened 2021-05-07
Avoid panicking while instrumenting the same DB driver twice

andrewslotin:
This PR addresses an issue when a database driver instrumentation with `instana.InstrumentSQLDriver()` may panic in case the same driver is instrumented twice. The panic is caused by an attempt to register a driver with the same name via `sql.Register()` and generally should not be the case, however we would like to avoid crashing the service in case of double instrumentation.


andriikushch (approved):
Looks good to me. 👍

PR #226, opened 2021-05-06
Direct lambda invocation add new span type

andriikushch:
This PR adds support for tracing the `Invoke` AWS API call. It has 3 logical parts. They are isolated in initial commits. 

- Adding registered type in the root project.
- Instrumenting SDK to inject tracing context into `Client Context` during `Invoke` call.
- Add support for the `Invoke` trigger in the lambda instrumentation, while handling the incoming call.

This PR is still missing the documentation and is in WIP. 🚧  
Any early feedback is more than welcome.

PR #228, opened 2021-05-21
add new span type for AWS Invoke

andriikushch:
This PR contains a definition of the new span for AWS Invoke and the new version number. It was created based on https://github.com/instana/go-sensor/pull/226.

### DIFF
--- a/registered_span.go
+++ b/registered_span.go
@@ -42,6 +42,8 @@ const (
 	AWSSNSSpanType = RegisteredSpanType("sns")
 	// AWS DynamoDB client span
 	AWSDynamoDBSpanType = RegisteredSpanType("dynamodb")
+	// AWS Lambda invoke span
+	AWSLambdaInvokeSpanType = RegisteredSpanType("aws.lambda.invoke")
 )
 
 // RegisteredSpanType represents the span type supported by Instana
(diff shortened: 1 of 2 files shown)


registered_span.go
```diff
@@ -1054,3 +1064,55 @@ func NewAWSDynamoDBSpanTags(span *spanS) AWSDynamoDBSpanTags {
 
 	return tags
 }
+
+// AWSInvokeSpanTags contains fields within the `aws.lambda.invoke` section of an OT span document
+type AWSInvokeSpanTags struct {
+	// FunctionName is a name of the function which is invoked
+	FunctionName string `json:"function"`
+	// InvocationType if equal to `Event`, means it is an async invocation
```
andrewslotin, line 1072: What are the other possible values here? If it's only to distinguish between sync/async, should we use a boolean instead?

version.go
```diff
@@ -4,4 +4,4 @@
 package instana
 
 // Version is the version of Instana sensor
-const Version = "1.28.1"
+const Version = "1.29.0"
```
andrewslotin, line 7: So far we've been bumping the version by pushing the commit directly to `master` before the release. I'm fine with either though.

andrewslotin (commented):
Overall LGTM. Let's clarify the invocation type value type and we're good

registered_span.go
andriikushch: There are three options (Event | RequestResponse | DryRun). These names are from the AWS documentation and they are shown in UI. This why I think it makes sense to have a string type here.

version.go
andriikushch: I agree, that it should not make a big difference.

andrewslotin (approved):

andrewslotin (approved):

andrewslotin (approved):

PR #230, opened 2021-05-21
add instrumentation for direct lambda invoke call

andriikushch:
Add instrumentation for direct lambda invoke call.


instrumentation/instaawssdk/instaawssdk.go
```diff
@@ -58,11 +64,13 @@ func InstrumentSession(sess *session.Session, sensor *instana.Sensor) {
...
 		case dynamodb.ServiceName:
 			FinalizeDynamoDBSpan(req)
+		case lambda.ServiceName:
+			FinalizeInvokeLambdaSpan(req)
 		}
 	})
 }
 
-func injectTraceContext(sp opentracing.Span, req *request.Request) {
+func injectTraceContext(logger instana.LeveledLogger, sp opentracing.Span, req *request.Request) {
```
andrewslotin, line 73: It's a matter of taste, but can we put auxiliary dependency to the end of parameter list?

andrewslotin (changes requested):
LGTM, just one minor comment.

andrewslotin (approved):

instrumentation/instaawssdk/instaawssdk.go
andriikushch: changed in next commit

PR #231, opened 2021-06-01
Add gin instrumentation

andriikushch:
This PR adds support for a gin framework
-----

1.<s>Functionality from TracingHandlerFunc was extracted to the new HttpEntrySpan type.</s>
2.<s>HttpEntrySpan</s> **TracingHandlerFunc** was used to implement gin instrumentation.
3. Gin framework provides method `Use` that inserts custom middleware, it appends handler at the end of its handlers. In this case, all middleware in the chain will not be traced. This is a reason why it was not used in this instrumentation.
4. Current instrumentation adds new middleware as a first handler that will init span, inject tracing context, and takes care of the correct span finishing.
5. Benchmark test was added for TracingHandlerFunc, to ensure that refactoring does not reduce performance.
6. Example for the gin instrumentation was added.
7. In UI gin tracing will look like regular HTTP traffic.

Important
-----

1. Before release, this PR must be divided into three parts:
- Change to the main module
- New instrumentation
- Example
2. Current codebase can be executed locally, because of go.mod with the local `replace` statement.

Architecture
-----

Gin engine has internally a slice with handlers, that it uses sequentially to process the incoming requests.
This instrumentation adds to the beginning of this slice an extra handler (middleware), that will be responsible for injecting/extracting tracing context.

**UPD:**

```
For example, after `gin.Default()` handlers slice will look like:
┌─────────────────────────────────────────────┐
│ Handlers                                    │
│                                             │
│ ┌──────────┐  context.Next() ┌───────────┐  │
│ │ Logger() ├─────────────────► Recovery()│  │
│ └──────────┘                 └───────────┘  │
│                                             │
└─────────────────────────────────────────────┘
After `instagin.AddMiddleware(sensor, engine)`, it will be transformed to:
┌──────────────────────────────────────────────────────────────────┐
│ Handlers                                                         │
│ ┌─────────────────────┐                                          │
│ │ TracingHandlerFunc  │                                          │
│ │     ...             │                                          │
│ │ ┌───────────────────┤                                          │
│ │ │ Replace request   │  ┌─────────┐ context.Next()┌───────────┐ │
│ │ │ context.Next()    ├──► Logger()├───────────────► Recovery()│ │
│ │ │ Write Status      │  └─────────┘               └───────────┘ │
│ │ └───────────────────┤                                          │
│ │     ...             │                                          │
│ └─────────────────────┘                                          │
│                                                                  │
└──────────────────────────────────────────────────────────────────┘
```


example/gin/README.md
```diff
@@ -0,0 +1,29 @@
+An Instrumented Server gin Example
+==========================
+
+An example of usage of Instana tracer in the instrumented gin application. It demonstrates how to use `https://github.com/instana/go-sensor/tree/master/instrumentation/instagin`.
```
andrewslotin, line 4:
```suggestion
An example of usage of Instana tracer to instrumenta a Gin application with `https://github.com/instana/go-sensor/tree/master/instrumentation/instagin`.
```

example/gin/README.md
```diff
@@ -0,0 +1,29 @@
...
+In case when the port is already in use, please choose another one.
+
+```
+  -address string
+        address to use by an example (default "localhost")
+  -port string
+        port to use by an example (default "8881")
+```
+
+There will be two endpoints exposed.
```
andrewslotin, line 24:
```suggestion
There will be two endpoints exposed:
```

example/gin/go.mod
```diff
@@ -0,0 +1,14 @@
...
+go 1.11
+
+require (
+	github.com/gin-gonic/gin v1.7.2
+
+	github.com/instana/go-sensor v1.29.0
+	github.com/instana/go-sensor/instrumentation/instagin v1.0.0
+)
+
+// todo: remove before release
```
andrewslotin, line 12: Let's not forget to do this 😉

example/gin/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/instana/go-sensor/instrumentation/instagin"
+)
+
+const (
+	defaultPort    = "8881"
+	defaultAddress = "localhost"
+)
+
+func main() {
+	port := flag.String("port", defaultPort, "port to use by an example")
```
andrewslotin, line 22: A minor note: other examples have a slightly different way to specify the listen addr. I'd suggest we follow the same pattern: https://github.com/instana/go-sensor/blob/master/example/http-database-greeter/main.go#L35

example/gin/main.go
```diff
@@ -0,0 +1,52 @@
...
+	engine.GET("/myendpoint", func(c *gin.Context) {
+		c.JSON(200, gin.H{
+			"message": "pong",
+		})
+	})
+
+	// use group: v1
+	v1 := engine.Group("/v1")
+
+	{
```
andrewslotin, line 43: Does Gin require to use a separate block to set up the handlers? Otherwise I'd suggest we omit it to simplify the example.

instrumentation/instagin/README.md
```diff
@@ -0,0 +1,40 @@
+Instana instrumentation for gin framework
+=============================================
+
+This module contains middleware to instrument gin framework applications.
```
andrewslotin, line 4:
```suggestion
This module contains middleware for instrumenting HTTP services written with [github.com/gin-gonic/gin](https://github.com/gin-gonic/gin).
```

instrumentation/instagin/go.mod
```diff
@@ -0,0 +1,13 @@
...
+go 1.11
+
+require (
+	github.com/gin-gonic/gin v1.7.2
+	github.com/instana/go-sensor v1.29.0
+	github.com/instana/testify v1.6.2-0.20200721153833-94b1851f4d65
+	github.com/opentracing/opentracing-go v1.2.0
+)
+
+// todo: remove before release
```
andrewslotin, line 12: Let's not forget to remove it 😉

instrumentation/instagin/handler.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"net/http"
+	"reflect"
+
+	"github.com/gin-gonic/gin"
+	instana "github.com/instana/go-sensor"
+)
+
+// AddMiddleware adds Middleware to the gin Handlers list. Unlike method Use from the current gin API, it adds
```
andrewslotin, line 16:
```suggestion
// AddMiddleware adds the tracing middleware to the list of Gin handlers. Unlike the gin.Use method, it puts the middleware
```

instrumentation/instagin/handler.go
```diff
@@ -0,0 +1,55 @@
...
+// to the beginning of the list. It will allows to trace all the default handlers added during gin.Default() call.
```
andrewslotin, line 17:
```suggestion
// to the beginning of the list to allow tracing default handlers added by the gin.Default() call.
```

instrumentation/instagin/handler.go
```diff
@@ -0,0 +1,55 @@
...
+	// trigger engine.rebuild404Handlers and engine.rebuild405Handlers
+	engine.Use()
+}
+
+type statusWriter interface {
+	SetStatus(status int)
+}
+
+// middleware wraps gin's handlers execution. Adds tracing context and handles entry span.
+var middleware = func(sensor *instana.Sensor) gin.HandlerFunc {
```
andrewslotin, line 31: I'm curious what's the benefit of this node-style declaration. Would the regular `func middleware(sensor *instana.Sensor) gin.HandlerFunc` work?

instrumentation/instagin/handler.go
```diff
@@ -0,0 +1,55 @@
...
+	return func(gc *gin.Context) {
+		instana.TracingHandlerFunc(sensor, "", func(writer http.ResponseWriter, request *http.Request) {
```
andrewslotin, line 33: Can we extract the path template that has been matched here and pass it to the `instana.TracingHandlerFunc()`?

instrumentation/instagin/handler.go
```diff
@@ -0,0 +1,55 @@
...
+			// set status from gc.Writer to instana.statusCodeRecorder which is used by instana.TracingHandlerFunc
+			if v, ok := writer.(statusWriter); ok {
+				v.SetStatus(gc.Writer.Status())
+			}
+		})(gc.Writer, gc.Request)
+	}
+}
+
+// tryFindAndRemove tries to find a previously registered middleware and remove it from the handlers list.
+// This function not necessarily is able to find duplicates. See documentation for a Pointer() method.
```
andrewslotin, line 46: > This function not necessarily is able to find duplicates

So it does not guarantee that we don't double-instrument the handler? Shall we remove it then to not to do deal with reflection?

instrumentation/instagin/handler_test.go
```diff
@@ -0,0 +1,173 @@
...
+	recorder := instana.NewTestRecorder()
+	tracer := instana.NewTracerWithEverything(
+		&instana.Options{Tracer: instana.TracerOptions{
+			CollectableHTTPHeaders: []string{"x-custom-header-1", "x-custom-header-2"},
+		},
+		}, recorder)
+
+	sensor := instana.NewSensorWithTracer(tracer)
+
+	engine := gin.Default()
```
andrewslotin, line 107: A minor note: you might consider extracting a helper function that initializes an instrumented Gin engine.

instrumentation/instagin/handler_test.go
```diff
@@ -0,0 +1,173 @@
...
+	assert.Equal(t, entrySpan.TraceID, interSpan.TraceID)
+	assert.Equal(t, entrySpan.SpanID, interSpan.ParentID)
+
+	assert.Equal(t, traceIDHeader, instana.FormatID(entrySpan.TraceID))
+	assert.Equal(t, spanIDHeader, instana.FormatID(entrySpan.ParentID))
+
+	require.IsType(t, instana.HTTPSpanData{}, entrySpan.Data)
+	entrySpanData := entrySpan.Data.(instana.HTTPSpanData)
+
+	require.IsType(t, instana.SDKSpanData{}, interSpan.Data)
```
andrewslotin, line 149: I'd suggest we don't test the test handler, as it has nothing to do with our instrumentation besides context propagation. Check that it's a child of the span created by instagin would be enough.

instrumentation_http.go
```diff
@@ -218,6 +218,10 @@ type statusCodeRecorder struct {
 	Status int
 }
 
+func (rec *statusCodeRecorder) SetStatus(status int) {
+	rec.Status = status
+}
+
 func (rec *statusCodeRecorder) WriteHeader(status int) {
 	rec.Status = status
```
andrewslotin, line 226: Shall we `SetStatus()` here?

andrewslotin (changes requested):
Overall looks very good, added some minor notes that I suggest we address before the release.

instrumentation/instagin/handler.go
andriikushch: I think this function will do its job in most cases. Nevertheless, I will remove this functionality for the reason you mentioned.

instrumentation/instagin/handler_test.go
andriikushch: I was thinking about it and found a few ways to create an engine object. Users can create it with default parameters (method **Default**), without parameters (method **New**) or just create an object with &gin.Engine. Therefore my idea was to use a unified way to cover all these cases with **AddMiddleware**. In my opinion, it will simplify usage.

instrumentation/instagin/handler_test.go
andriikushch: I will simplify test

instrumentation_http.go
andriikushch: In this case, it is a matter of taste. Regarding the `SetStatus` internals and assuming that it will be inlined here. So I am fine with both ways of using it and will update `WriteHeader` and `Write` functions.

andrewslotin: Let's update them all 👍  In this case it's unlikely we add more functionality to the setter, but generally speaking I think it's good to use a consistent way of accessing the value.

instrumentation/instagin/handler_test.go
andrewslotin: Right, my idea was to make it a private test helper and not generalize it too much. But it's up to you!

andriikushch: I read the wrong filename (did not recognize that it is about _test) 😄  I will check it once. Good catch 👍

example/gin/main.go
andriikushch: I will change and create a reminder for grpc-client-server example.

instrumentation/instagin/handler.go
andriikushch: I will put

example/gin/go.mod
andriikushch: 👍

andriikushch:
This PR is divided into three:

1. Main module https://github.com/instana/go-sensor/pull/232
2. Example https://github.com/instana/go-sensor/pull/233
3. Gin instrumentation https://github.com/instana/go-sensor/pull/234

example/gin/main.go
andriikushch: It does not. I will change it in child PR.

instrumentation/instagin/handler.go
andriikushch: Done in child PR.

instrumentation/instagin/handler.go
andriikushch: added to child PR

instrumentation/instagin/handler_test.go
andriikushch: done in child PR

instrumentation/instagin/go.mod
andriikushch: done

instrumentation/instagin/handler.go
andriikushch: It was used for deduplication purposes before. Now it can be done "regular" way.

andriikushch:
This PR is divided into three:

Main module #232
Example #233
Gin instrumentation #234

PR #234, opened 2021-06-08
add gin instrumentation

andriikushch:
Is a part of https://github.com/instana/go-sensor/pull/231. Contains gin instrumentation.


andriikushch:
JFYI @andrewslotin

instrumentation/instagin/README.md
```diff
@@ -0,0 +1,40 @@
+Instana instrumentation for gin framework
+=============================================
+
+This module contains middleware to instrument gin framework applications.
```
andrewslotin, line 4: We usually link the instrumented package to provide context
```suggestion
This module contains middleware to instrument HTTP services written with [github.com/gin-gonic/gin](https://github.com/gin-gonic/gin).
```

instrumentation/instagin/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/instana/go-sensor/instrumentation/instagin
+
+go 1.11
+
+require (
+	github.com/gin-gonic/gin v1.7.2
+	github.com/instana/go-sensor v1.29.1-0.20210608074124-c12a7d59b2d4
```
andrewslotin, line 7: Let's not forget to update it to a released version before merging.

instrumentation/instagin/handler_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/opentracing/opentracing-go"
+
+	"github.com/instana/testify/assert"
+
+	"github.com/gin-gonic/gin"
+	instana "github.com/instana/go-sensor"
+	"github.com/instana/go-sensor/instrumentation/instagin"
+)
+
+func TestAddMiddleware(t *testing.T) {
```
andrewslotin, line 24: Is this test still needed after we removed the `findAndReplace()`? I'd rather suggest that we check if an instrumented handler collects expected information in a span, i.e. the path template, URL, params, etc.

andrewslotin (changes requested):
Overall LGTM. I like the simplicity of implementation 👍  Let's add a test to check the span content and we're good to go

instrumentation/instagin/go.mod
andriikushch: 👍

instrumentation/instagin/handler_test.go
andriikushch: This test, checks that middleware added as a first handler, no matter if there are some middleware already added. I will add a test to check data, as you suggested.

andrewslotin: I see 👍  I'm fine with current implementation, but as a note you might have added a middleware and check that there is a span in request context there, since this is wha we're trying to achieve. This way we're not dependent on Gin implementation

instrumentation/instagin/handler_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	assert.Len(t, engine.Handlers, expectedHandlersAmount)
+
+	// check that middleware was added as a first handler
+	assert.NotEqual(t, handlerN1Pointer, reflect.ValueOf(engine.Handlers[0]).Pointer())
+	assert.Equal(t, handlerN1Pointer, reflect.ValueOf(engine.Handlers[1]).Pointer())
+	assert.Equal(t, handlerN2Pointer, reflect.ValueOf(engine.Handlers[2]).Pointer())
+}
+
+func TestResponseHeadersCollecting(t *testing.T) {
```
andrewslotin, line 45: Nitpicking: this test ensures propagation, not header collection 😉 The header collection would be populating the [`CollectableHTTPHeaders`](https://github.com/instana/go-sensor#capturing-custom-http-headers) option and checking that the entry span has corresponding values. You might want to do this together with checking the span content above.

instrumentation/instagin/handler_test.go
andriikushch: done.

andriikushch:
@andrewslotin I added suggested assertions and changed a declaration of the _func middleware(sensor *instana.Sensor)_ .

andrewslotin (approved):
👍

PR #239, opened 2021-07-05
Do not send empty tracestate headers downstream

andrewslotin:
The `tracestate` is an optional header to propagate vendor states as a part of W3C trace context. In case Instana trace is suppressed either by `X-Instana-L: 0` header or corresponding flag in `traceparent`, the tracer won't be adding any data to the `tracestate` header. In this case, if a trace context did not contain any other vendor data before, we'd send a header with an empty value. This PR addresses this behavior by skipping adding the `tracestate` header is the state is empty.


andriikushch:
Looks good! 👍

andriikushch (approved):

PR #241, opened 2021-07-19
Add logging span to the list of registered spans

andrewslotin:
This PR adds a new registered span type intended for logger instrumentation.

### DIFF
--- a/registered_span.go
+++ b/registered_span.go
@@ -44,6 +44,8 @@ const (
 	AWSDynamoDBSpanType = RegisteredSpanType("dynamodb")
 	// AWS Lambda invoke span
 	AWSLambdaInvokeSpanType = RegisteredSpanType("aws.lambda.invoke")
+	// Logging span
+	LogSpanType = RegisteredSpanType("log.go")
 )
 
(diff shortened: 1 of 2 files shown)


registered_span.go
```diff
@@ -1116,3 +1127,52 @@ func newAWSLambdaInvokeSpanData(span *spanS) AWSLambdaInvokeSpanData {
...
 }
+
+// LogSpanData represents the `data` section of a logging span
+type LogSpanData struct {
+	SpanData
+	Tags LogSpanTags `json:"log"`
+}
+
+// NewLogSpanData initializes a new logging span data from tracer span
+func NewLogSpanData(span *spanS) LogSpanData {
```
andriikushch, line 1138: What do you think about renaming it to `newLogSpanData`, because we can not call `NewLogSpanData` from another package, due to its "private" argument `span *spanS`?

andriikushch (approved):
Please check my comment. 

Overall looks good to me! Good job! 👍 Feel free to merge.

registered_span.go
andrewslotin: Makes sense, will do 👍

PR #242, opened 2021-07-26
Add missing copyright headers

andrewslotin:
This PR adds missing copyright headers as well as `Makefile` target to ensure that we won't miss them in future.


andriikushch (approved):
Good catch! 👍

andriikushch (approved):
Looks good! 👍

PR #243, opened 2021-07-27
Add OpenTracing log support

andrewslotin:
This PR adds support for forwarding the log messages created via the OpenTracing span API to Instana backend, while retiring the old and unsupported way of sending them as a part of SDK span payloads. Once merged, the log records that contain fields with either `"error"` or `"error.object"` keys will be forwarded as log spans and appear in the "Logs" view for the trace.

### DIFF
--- a/span.go
+++ b/span.go
@@ -4,6 +4,7 @@
 package instana
 
 import (
+	"strings"
 	"sync"
 	"time"
 
@@ -72,6 +73,7 @@ func (r *spanS) FinishWithOptions(opts ot.FinishOptions) {
 	r.Duration = duration
 	if !r.context.Suppressed {
 		r.tracer.recorder.RecordSpan(r)
+		r.sendOpenTracingLogRecords()
 	}
 }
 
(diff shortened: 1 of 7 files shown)


span.go
```diff
@@ -181,3 +184,44 @@ func (r *spanS) SetTag(key string, value interface{}) ot.Span {
...
+			"log.message": strings.TrimSpace(buf.String()),
+		},
+	).FinishWithOptions(
+		ot.FinishOptions{
+			FinishTime: lr.Timestamp,
+		},
+	)
+}
+
+func isErrorLogField(lf otlog.Field) bool {
```
andriikushch, line 225:
```suggestion
func (r *spanS) isErrorLogField(lf otlog.Field) bool {
```
This function is used only inside spanS, what do you think if we will make a method of spanS?

andriikushch (approved):
Good job 👍  Looks good! Please check my comment if it makes sense to you. Feel free to merge it.

span.go
andrewslotin: I see your point to use the receiver as a namespace, however this helper function is relevant to any `opentracing.Field` and I believe there is no need to restrict its usage to our spans only. Do you see any disadvantages in this approach?

andriikushch: In this particular case, I don't think there are any. It is more about "cohesion" and responsibility in general. Having a lot of helpers function may lead to creating the packages/modules like "util", which is often considered a not best practice. Nevertheless, I see your point, and being pragmatic keeping as it is is absolutely fine for me. 👍

PR #246, opened 2021-08-16
Add issue templates

andrewslotin:
This PR adds issue templates similar to the ones that are used in [Instana ruby-sensor repository](https://github.com/instana/ruby-sensor/issues/new/choose).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -38,10 +38,10 @@ body:
     validations:
       required: true
   - type: textarea
-    id: go-version
+    id: goenv
     attributes:
-      label: Go Version
-      description: What version of Go was the application compiled with when it encountered this bug?
+      label: go env
+      description: Please paste the output of `go env` from the machine that has been used to build the application.
       render: shell
     validations:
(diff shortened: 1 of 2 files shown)


andriikushch (approved):
Looks good to me! 👍  Good job @andrewslotin! 

I have two small questions: 

1. Can we test it before merge?
2. What do you think about adding a field for a `go evn` output? Will it provide valuable information for bug fixing?

andrewslotin:
> Can we test it before merge?

I just used the templates from our Ruby sensor, so I expect it to work the same way. It looks like GH considers creating custom templates to be an advanced feature, at least I could not find a UI for it.

> What do you think about adding a field for a `go env` output? Will it provide valuable information for bug fixing?

I think this is a good idea, we could do this instead of asking for Go version. [github.com/golang/go](https://github.com/golang/go) does something similar

andrewslotin:
@andriikushch, I found that you can see the rendered view by opening the [template file](https://github.com/instana/go-sensor/blob/add_issue_templates/.github/ISSUE_TEMPLATE/bug.yml) on GH.

hmadison (approved):
👏

PR #247, opened 2021-08-17
Add gorilla mux instrumentation (instrumentation part)

andriikushch:
Instrumentation part of https://github.com/instana/go-sensor/pull/245


instrumentation/instamux/handler.go
```diff
@@ -0,0 +1,28 @@
...
+	instana "github.com/instana/go-sensor"
+)
+
+func AddMiddleware(sensor *instana.Sensor, router *mux.Router) {
+	router.Use(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			pathTemplate, err := mux.CurrentRoute(r).GetPathTemplate()
+			if err != nil {
+				pathTemplate = ""
+			}
```
andrewslotin, line 21: What could be the error here? Shall we add a debug log?

andrewslotin (approved):
LGTM! I've added a suggestion to log the path template retrieval error for your consideration

instrumentation/instamux/handler.go
andriikushch: There might be errors like when:
1. route does not define a path
2. when error resulted from building a route

I will add a logger, I do not see any reason not to have it. 👍

andriikushch: @andrewslotin Here it is https://github.com/instana/go-sensor/pull/247/commits/b3d9874b1042d1bd65ee521b5599f1188084a0f8

PR #248, opened 2021-08-17
Add gorilla mux instrumentation example

andriikushch:
Instrumentation part for #245 

**Build failed, because instrumentation need to be released first**


andrewslotin (approved):
Excellent! Let's get this merged once https://github.com/instana/go-sensor/pull/247 is done

PR #250, opened 2021-08-24
Instrument github.com/julienschmidt/httprouter

andrewslotin:
This PR adds an instrumentation module for HTTP servers written with `github.com/julienschmidt/httprouter` router.


andrewslotin:
@andriikushch, thanks for the review. Regarding your suggestion indeed, this was my first idea, however obtaining the path template in `ServeHTTP()` was the reason why I decided against it. The `httprouter.Router` parses the template path upon registration and creates a radix tree that is then used for handler lookup. The used tree path is not available to the handler, so we end up with the raw request path and a set of parameters only.

andriikushch (approved):
Looks good! 👍

PR #254, opened 2021-09-23
add echo framework instrumentation

andriikushch:
This PR adds instrumentation for the Echo(v4) framework. It is missing an example, it will be added with the next PR (therefore this will have one broken link).

It uses the same approach as was used for other frameworks.


instrumentation/instaecho/handler.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"net/http"
+
+	instana "github.com/instana/go-sensor"
+	"github.com/labstack/echo/v4"
+)
+
+// AddMiddleware adds the tracing middleware to the list of Echo handlers.
+func AddMiddleware(sensor *instana.Sensor, engine *echo.Echo) {
```
andrewslotin, line 16: What is the purpose of this function? Shall we expose the `middleware()` and let users call `engine.Use()` themselves?

instrumentation/instaecho/README.md
```diff
@@ -0,0 +1,40 @@
...
+-----
+
+```go
+// init Echo
+e := echo.New()
+
+// create a sensor
+sensor := instana.NewSensor("echo-sensor")
+
+// add middleware to the Echo's handlers
```
andrewslotin, line 28: We should emphasize that middleware should be added before declaring any handlers:
> Middleware registered using `Echo#Use()` is only executed for paths which are registered after `Echo#Use()` has been called.

[Source](https://echo.labstack.com/middleware/)

andrewslotin (changes requested):
Considering that our tracing middleware should be registered before any handlers, would it make sense instead of `instaecho.AddMiddleware()` provide `instaecho.New()` that would create a new Echo engine and register our middleware immediately?

instrumentation/instaecho/README.md
andriikushch: Makes sense! Good catch! 👍

instrumentation/instaecho/handler.go
andriikushch: It was done to have a similar interface as we already have for a gin framework. But I will agree, here it looks artificial. I will remove it. 
 

> Considering that our tracing middleware should be registered before any handlers, would it make sense instead of instaecho.AddMiddleware() provide instaecho.New() that would create a new Echo engine and register our middleware immediately?

I think you made a good suggestion with an `instaecho.New()` method. 👍 As far as there is only one constructor, let's wrap it.

andriikushch:
> Considering that our tracing middleware should be registered before any handlers, would it make sense instead of instaecho.AddMiddleware() provide instaecho.New() that would create a new Echo engine and register our middleware immediately?

It sounds like a good idea 👍 , please check the last commit.

instrumentation/instaecho/README.md
```diff
@@ -0,0 +1,37 @@
+Instana instrumentation for echo framework
```
andrewslotin, line 1:
```suggestion
Instana instrumentation for Echo framework
```

instrumentation/instaecho/README.md
```diff
@@ -0,0 +1,37 @@
...
+```go
+// create a sensor
+sensor := instana.NewSensor("echo-sensor")
+
+// init instrumented Echo
+e := instaecho.New(sensor)
+
+// define API
+e.GET("/foo", func(c echo.Context) error {...})
+...
```
andrewslotin, line 30: Just to make sure this snippet is a valid Go code
```suggestion
e.GET("/foo", func(c echo.Context) error { /* ... */ })
// ...
```

andrewslotin (approved):
Looks good! I made two minor formatting suggestions to the README, feel free to merge.

PR #262, opened 2021-10-13
Detect "cold" Lambda invocation

andriikushch:


andriikushch:
@andrewslotin, I've changed it to use `sync.Once`. Looks good to me. 👍

PR #272, opened 2021-11-01
Add a method to report agent client readiness

andrewslotin:
This PR introduces a new API method `instana.Ready()` that reports whether the Instana collector is ready to collect and send trace data.


andriikushch (approved):
👍

PR #273, opened 2021-11-01
Expose an API method to flush buffered data to the agent

andrewslotin:
This PR introduces the `instana.Flush()` to flush all buffered data to the agent. This method is intended to gracefully shut down the service and ensure that all trace data has been sent before exit. It's not recommended to use it to interfere with data transmission cycle, as it's not guaranteed to leave the agent client in operational state.


andriikushch (approved):
Looks good!

PR #274, opened 2021-11-05
Forward warning logs collected via the OpenTracing API to the agent

andrewslotin:
This PR updates Go collector to send the OpenTracing warning log records to the agent. A record is considered as a warning if it contains at least one field with `"warn"` as a key. Such logs are send with `WARN` level.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -24,6 +24,12 @@ const DefaultPrefix = "instana: "
 // Level defines the minimum logging level for logger.Log
 type Level uint8
 
+// Less returns whether the log level is less than the given one in logical order:
+// ErrorLevel > WarnLevel > InfoLevel > DebugLevel
+func (lvl Level) Less(other Level) bool {
+	return uint8(lvl) > uint8(other)
+}
+
 // String returns the log line label for this level
 func (lvl Level) String() string {
 	switch lvl {
--- a/span.go
+++ b/span.go
@@ -9,9 +9,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/instana/go-sensor/logger"
 	ot "github.com/opentracing/opentracing-go"
 	otlog "github.com/opentracing/opentracing-go/log"
 )
+
+const minSpanLogLevel = logger.ErrorLevel
 
 type spanS struct {
 	Service     string
@@ -115,7 +118,7 @@ func (r *spanS) LogFields(fields ...otlog.Field) {
 
 	for _, v := range fields {
 		// If this tag indicates an error, increase the error count
-		if isErrorLogField(v) {
+		if openTracingLogFieldLevel(v) == logger.ErrorLevel {
 			r.ErrorCount++
 		}
 	}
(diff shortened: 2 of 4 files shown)


span.go
```diff
@@ -189,16 +192,17 @@ func (r *spanS) Tracer() ot.Tracer {
...
-			}
-		}
+		r.sendOpenTracingLogRecord(lr)
 	}
 }
 
 func (r *spanS) sendOpenTracingLogRecord(lr ot.LogRecord) {
+	lvl := openTracingLogRecordLevel(lr)
+
+	if lvl > minSpanLogLevel {
```
andriikushch, line 202: @andrewslotin What do you think: based on the description of this PR, does it makes sense instead of returning here to do filtering based on the minimal level we want to return?

span.go
```diff
@@ -222,6 +226,25 @@ func (r *spanS) sendOpenTracingLogRecord(lr ot.LogRecord) {
 	)
 }
 
-func isErrorLogField(lf otlog.Field) bool {
-	return lf.Key() == "error" || lf.Key() == "error.object"
+func openTracingLogRecordLevel(lr ot.LogRecord) logger.Level {
```
andriikushch, line 229: What do you think if we will rename this function to something like `highestLogRecordLevel` or similar?

andriikushch (changes requested):
It looks good to me, please check my comments and then I think we can merge it.

span.go
```diff
@@ -222,6 +226,25 @@ func (r *spanS) sendOpenTracingLogRecord(lr ot.LogRecord) {
...
+	return highestLvl
+}
+
+func openTracingLogFieldLevel(lf otlog.Field) logger.Level {
+	switch lf.Key() {
+	case "error", "error.object":
+		return logger.ErrorLevel
+	case "warn":
+		return logger.WarnLevel
+	default:
```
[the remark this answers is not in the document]
andrewslotin: We only collect logs with level WARN and above in other tracers, so INFO and others are omitted intentionally.

span.go
andrewslotin: I'm not sure I follow, could you elaborate what did you meant by "filtering based on the minimal level we want to return"?

span.go
andrewslotin: Good point, that would be more descriptive method name 👍

span.go
andriikushch: Let me clarify it a little bit 😃 

If I follow logic correct: we want to send logs with level warn or error. 

I see in code we iterate over `lr.Fields` to find what is the highest log level is in the `LogRecord`. After, if this level is higher than `WarnLevel` we do not send this record.

From the code perspective, it seems there might be multiple fields within `lr` with different levels. With the current implementation, if any has a level higher than `warn` whole `LogRecord` won't be sent. 

So by filtering, I mean ignoring (filtering out) only fields where `lf.Key` is not an `error` or `warn`, but not whole the `LogRecord`. 

Does it make sense?

andrewslotin: Okay, I see your point now. The thing is that the logic here is supposed to be quite the opposite: if we find at least one field with key `error` or `warn`, the whole log record is considered to be of this level and thereby needs to be sent to the agent.

The problem is that log levels in the `logger` package are represented by integer numbers with `0` being the highest level `ERROR`, so we need to invert the comparison to figure out whether we should discard this record or not. I was thinking of introducing a separate type for log levels here with a different set of values, so that `ERROR` would be greater than `WARN`, but found it to be unnecessary.

Do you think it would help if I add a comment here?

andriikushch: Ok, got it. I think a tiny comment will be helpful there.

andriikushch (approved):
LGTM 👍

span.go
andrewslotin: I ended up with adding `(logger.Level).Less()`, does this code make more sense now?

logger/logger.go
```diff
@@ -24,6 +24,12 @@ const DefaultPrefix = "instana: "
 // Level defines the minimum logging level for logger.Log
 type Level uint8
 
+// Less returns whether the log level is less than the given one in logical order:
+// ErrorLevel > WarnLevel > InfoLevel > DebugLevel
+func (lvl Level) Less(other Level) bool {
+	return uint8(lvl) > uint8(other)
```
andriikushch, line 30: I think we do not require type cast here. What do you think?
```suggestion
	return lvl > other
```

andriikushch (approved):
Looks good 👍

logger/logger.go
andrewslotin: I agree that it's redundant from the code perspective, however I think it increases readability explicitly telling that we compare integers there.

PR #275, opened 2021-11-08
Add instrumentation for github.com/sirupsen/logrus

andrewslotin:
This PR adds a new instrumentation package to instrument `github.com/sirupsen/logrus` logger with Instana. An instrumented logger collects and sends any error or warning log records to the agent if the call is traced.


instrumentation/instalogrus/instalogrus.go
```diff
@@ -0,0 +1,71 @@
...
+		h.sensor.Logger().Debug("ignoring logrus.Entry without context.Context")
+		return nil
+	}
+
+	parent, ok := instana.SpanFromContext(entry.Context)
+	if !ok {
+		return nil
+	}
+
+	msg, err := entry.Bytes()
```
andriikushch, line 40: I think it can be simplified. What do you think? 
```suggestion
	msg, err := entry.String()
```

instrumentation/instalogrus/instalogrus.go
```diff
@@ -0,0 +1,71 @@
...
+			"log.message": string(msg),
+		},
+	).FinishWithOptions(opentracing.FinishOptions{
+		FinishTime: entry.Time,
+	})
+
+	return nil
+}
+
+func convertLevel(lvl logrus.Level) string {
```
andriikushch, line 60: Not sure if this is better 😃  Just like an idea in order to not create extra mappings/constants, we can use something like :

```
func convertLevel(lvl logrus.Level) string {
	switch lvl {
	case logrus.ErrorLevel, logrus.WarnLevel, logrus.InfoLevel:
		return strings.ToUpper(lvl.String())
	default:
		return "DEBUG"
	}
}
```

andriikushch (approved):
Looks good 👍  Please check my comments and feel free to merge.

instrumentation/instalogrus/instalogrus.go
andrewslotin: I'm usually trying to avoid relying on constant values, since this would introduce an implicit coupling between packages, i.e. we won't be able to change let's say `WARN` to `WARNING`. I see your point that it would be shorter, but still think that an explicit version is less error-prone and more explanatory.

PR #276, opened 2021-11-12
draft for a version updater

andriikushch:
Here is a draft for a release hepler. 

To try: from cmd folder run `go run release.go --releaseType=bugfix`. It will (should 😃 ) not do any changes to the repository but modify a local file and print git commands which will be executed if `--execute` flag is provided.

I am thinking to create a separate repo for this, what do you think about @andrewslotin?


andrewslotin:
Looks interesting. We might also consider [GoReleaser](https://goreleaser.com/) for this

andrewslotin:
Okay, I've had another look at GoReleaser and it looks like it does not fit our purpose. If I get it right, it covers every aspect of release process, _except_ for bumping the version, which is exactly what we're trying to achieve here.

andrewslotin:
Replaced by #277

PR #279, opened 2021-11-25
Fix Makefile.release

andriikushch:
Makefile.release assumes that `version.go` has `var Version = ` statement, but it has `const Version = `


andrewslotin (approved):
Good catch!

PR #280, opened 2021-11-26
Increase a timeout for an agent

andriikushch:
During the announcement, the agent has to start multiple processes (sensors, profile and accomplish the discovery process) and usually it takes longer than 5 seconds. 

Experiments did show that having timeout >= 10s should be enough to announce with a single announcement. This change should improve discovery processes for a sensors version `>= 1.3.1`.

We can make this parameter configurable if we want. Nevertheless, increasing this value to `15` should improve the overall stability of the discovery process (has fewer retries and errors related to it).


andrewslotin:
This might slow down the data transmission since this timeout is used for every request made to the agent. Would it make sense to shorten the time between retries instead and add exponential backoff to make sure we don't DDoS the agent?

andriikushch:
> This might slow down the data transmission since this timeout is used for every request made to the agent. Would it make sense to shorten the time between retries instead and add exponential backoff to make sure we don't DDoS the agent?

I think it won't create any problem because the number of requests will stay ~ as was before. But I agree that changing it for every request looks "rude".

I made the following changes for now:

1. I created a method that can schedule "exponential" retries and integrated it in the places where it can be done easily. In my opinion, other places need more refactoring of the logic to build around "FSM". I don't want to make it part of this PR, because it will be hard to understand and review.
2. As it makes sense to have a longer timeout for an announce call, I changed it only for it. I used request with a timeout, I am not using `NewRequestWithContext` because it doesn't exist in older go versions. Also, I found that you can't make a particular request that has a longer timeout than a client has. For safety reasons, I still have a timeout for a client. So there can't be a request without timeout at all.
3. I also changed the number of retries from 2 to 3. So for "exponential" retries, it should be after 10, 20, 40 seconds.

agent.go
```diff
@@ -229,18 +235,20 @@ func (agent *agentS) head(url string) (string, error) {
 }
 
 func (agent *agentS) request(url string, method string, data interface{}) (string, error) {
-	return agent.fullRequestResponse(url, method, data, nil, "")
+	return agent.fullRequestResponse(nil, url, method, data, nil, "")
```
andrewslotin, line 238: Why not pass `context.Background()` instead of `nil`?

agent.go
```diff
@@ -251,13 +259,25 @@ func (agent *agentS) fullRequestResponse(url string, method string, data interfa
 		j, err = json.Marshal(data)
 	}
 
+	var cancel context.CancelFunc
+	if ctx == nil {
+		ctx, cancel = context.WithTimeout(context.Background(), agent.clientTimeout)
+		defer cancel()
+	}
+	if _, ok := ctx.Deadline(); !ok {
```
andrewslotin, line 267: I'm not sure I follow why client timeout is only used if there is no deadline set. Is this to not to override longer deadline? In this case I'd argue this is the right place to add this timeout to the context. Since it's the caller to decide which period of time to use, I'd suggest we set the timeout there.

fsm.go
```diff
@@ -67,6 +69,14 @@ func (r *fsmS) scheduleRetry(e *f.Event, cb func(e *f.Event)) {
 	}()
 }
 
+func (r *fsmS) scheduleRetryWithExponentialDelay(e *f.Event, cb func(e *f.Event), retryNumber int) {
+	r.timer = time.NewTimer(getRetryPeriodForIteration(retryNumber))
+	go func() {
+		<-r.timer.C
+		cb(e)
+	}()
+}
```
andrewslotin, line 78: A function that spawns a goroutine and return immediately are hard to test and they tend to be prone to leaking goroutines. I'd suggest we use a simple `time.Sleep()` here with a subsequent call to `cb()` and leave it up to the caller side to control whether to invoke this callback in a sync or async way.

fsm.go
```diff
@@ -237,3 +249,7 @@ func (r *fsmS) cpuSetFileContent(pid int) string {
 
 	return string(data)
 }
+
+func getRetryPeriodForIteration(retryNumber int) time.Duration {
```
andrewslotin, line 253: This function is very specific as suggested by its long name 😅 How about we allow callers to provide the initial delay as well? Then we could name it simply `expDelay()` or something like this. WDYT?

andrewslotin (changes requested):
I like the idea, but think we need to make some adjustments before merging it

agent.go
andriikushch: The idea was to ensure that in the case when the function is called with, for instance: "context.Background", set Deadline here. If ctx has already Deadline, keep it.

If the caller decided to set a specific Deadline it will be used.

agent.go
andriikushch: Actually no reason for that. It will be set inside a function anyway, but if looks better, I will add it her 👍

fsm.go
andriikushch: Absolutely agree, it was copy/paste of `scheduleRetry`. I will change it in a sec.

fsm.go
andriikushch: `expDelay` sounds good

agent.go
andrewslotin: I believe it's also a convention to not to pass `nil` context and use `context.Background()`/`context.TODO()` instead.

agent.go
andrewslotin: Alright, so what's the purpose of setting the timeout in this method? My current understanding is that this is some kind of default that should satisfy any use case, but I'd argue that this should be the biggest value out of announce and data transmission values, and not the smallest one, i.e. as per your other comment, there should be no problem with sending data using 15s timeout, but an announcement limited to 5s will likely to fail. So I'd suggest to use 15s as a default instead. This would also simplify the implementation here, since the shortest timeout will cancel the context even if the longer one has not been reached yet, meaning that you can always do
```go
ctx, cancel := context.WithTimeout(ctx, agent.announceTimeout)
```
and let the data transmission methods set their own shorter deadline.

agent.go
```diff
@@ -97,11 +99,15 @@ func newAgent(serviceName, host string, port int, logger LeveledLogger) *agentS
...
-		from:   &fromS{},
-		host:   host,
-		port:   strconv.Itoa(port),
-		client: &http.Client{Timeout: 5 * time.Second},
+		from:            &fromS{},
+		host:            host,
+		port:            strconv.Itoa(port),
+		clientTimeout:   5 * time.Second,
+		announceTimeout: 15 * time.Second,
+		client:          &http.Client{Timeout: maxClientTimeout},
```
andrewslotin, line 110: Apologies, I've missed this change during the last review. What's the reason behind it and where did 60s come from?

agent.go
andrewslotin: I believe this would make the logic clearer too: by default agent client gives up after 15s, but for data transmission requests we require shorter timeouts to not to overload the agent.

agent.go
andriikushch: The single reason is always to have at least some timeout. Because I believe by default there is no timeout at all.

andrewslotin: I see, shall we use 15s then and shorten it to 5s where appropriate? Wouldn't this simplify the logic?

andriikushch: I changed this now.

agent.go
andriikushch: I think you are right that code is less straightforward now 😃 . @andrewslotin please take a look at https://github.com/instana/go-sensor/pull/280/commits/0e369bc1fd05aa309579a30d09436e8eba0b2e7e. I tried to simplify and make it clearer. So now:

1. Client has a timeout (which is equal to announceTimeout which is the biggest we need so far). Before it was 5s and now it is a 15s.
2. We have 3 methods that call actual call (`fullRequestResponse` method):

```
func (agent *agentS) request(url string, method string, data interface{}) (string, error) {
	ctx, cancel := context.WithTimeout(context.Background(), agent.clientTimeout)
	defer cancel()
	return agent.fullRequestResponse(ctx, url, method, data, nil, "")
}

func (agent *agentS) announceRequest(url string, method string, data interface{}, ret *agentResponse) (string, error) {
	return agent.fullRequestResponse(context.Background(), url, method, data, ret, "")
}

func (agent *agentS) requestHeader(url string, method string, header string) (string, error) {
	ctx, cancel := context.WithTimeout(context.Background(), agent.clientTimeout)
	defer cancel()
	return agent.fullRequestResponse(ctx, url, method, nil, nil, header)
}
```

2 of them overwrite (reduce) timeout, so their behavior will be as before. And announce call will use a default one that is equal to 15s now.

agent.go
```diff
@@ -97,11 +99,14 @@ func newAgent(serviceName, host string, port int, logger LeveledLogger) *agentS
 
 	logger.Debug("initializing agent")
 
+	announceTimeout := 15 * time.Second
```
andrewslotin, line 102: I guess this can be inlined now, right? Shall we introduce constants for both timeouts?

fsm.go
```diff
@@ -237,3 +246,7 @@ func (r *fsmS) cpuSetFileContent(pid int) string {
 
 	return string(data)
 }
+
+func expDelay(retryNumber int) time.Duration {
+	return time.Duration(math.Pow(2, float64(retryNumber-1))) * exponentialRetryPeriodBase
```
andrewslotin, line 251: Just curious why do we always need to add 1 to the retry number when calling this function to subtract it later?

agent.go
```diff
@@ -229,18 +234,22 @@ func (agent *agentS) head(url string) (string, error) {
 }
 
 func (agent *agentS) request(url string, method string, data interface{}) (string, error) {
-	return agent.fullRequestResponse(url, method, data, nil, "")
+	ctx, cancel := context.WithTimeout(context.Background(), agent.clientTimeout)
```
andrewslotin, line 237: Since it's not clear from the method name that these methods are used for the data transmission cycle, shall we put a comment that explains why there is a different timeout here?

agent.go
```diff
@@ -229,18 +234,22 @@ func (agent *agentS) head(url string) (string, error) {
...
 
-func (agent *agentS) requestResponse(url string, method string, data interface{}, ret interface{}) (string, error) {
-	return agent.fullRequestResponse(url, method, data, ret, "")
+func (agent *agentS) announceRequest(url string, method string, data interface{}, ret *agentResponse) (string, error) {
+	return agent.fullRequestResponse(context.Background(), url, method, data, ret, "")
 }
 
 func (agent *agentS) requestHeader(url string, method string, header string) (string, error) {
-	return agent.fullRequestResponse(url, method, nil, nil, header)
+	ctx, cancel := context.WithTimeout(context.Background(), agent.clientTimeout)
```
andrewslotin, line 247: Same as above, shall we elaborate on using different timeout?

andrewslotin (approved):
Overall LGTM, I like how the logic became clearer now. I've put some minor suggestions, but generally speaking I think it's good to go

fsm.go
andriikushch: Actually good question :) I wanted to have a variable with a clear name - `retryNumber`. So for a caller, it makes sense: try the first time, second, etc. (2^0, 2^1, 2^2 ...) so I do minus 1. On the caller side, I have to calculate the number of current retry. In a way how "fsm" works now, without changing a logic I have to do `retryNumber := maximumRetries - r.retriesLeft + 1`. I plan to change this logic when will work on simplification/removal of "fsm". This is a plan.

agent.go
andriikushch: Agree

agent.go
andriikushch: 👍

PR #282, opened 2021-12-02
Make HTTP instrumentation wrapper implement `http.Hijacker` in case the `http.ResponseWriter` does

andrewslotin:
This PR updates the HTTP instrumentation wrapper to implement `http.Hijacker` whenever the underlying `http.ResponseWriter` implements this interface.

It also lifts the min Go version for `github.com/instana/go-sensor` because of type aliases support.

### DIFF
--- a/instrumentation_http.go
+++ b/instrumentation_http.go
@@ -4,8 +4,10 @@
 package instana
 
 import (
+	"bufio"
 	"context"
 	"mime/multipart"
+	"net"
 	"net/http"
 	"net/textproto"
 	"net/url"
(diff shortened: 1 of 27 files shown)


instrumentation_http.go
```diff
@@ -219,14 +221,31 @@ func RoundTripper(sensor *Sensor, original http.RoundTripper) http.RoundTripper
...
+}
+
+func wrapResponseWriter(w http.ResponseWriter) wrappedResponseWriter {
+	if _, ok := w.(http.Hijacker); ok {
+		return &statusCodeRecorderHTTP10{
+			ResponseWriter: w,
+		}
+	}
+
+	return &statusCodeRecorder{
```
andriikushch, line 236: What would be if we always return statusCodeRecorderHTTP10 from this method?

andriikushch (approved):

andriikushch:
Looks good 👍

PR #284, opened 2021-12-08
Add validation of W3C trace context header

andriikushch:
Add validation of W3C trace context header based on https://www.w3.org/TR/trace-context/#header-name spec


andrewslotin (commented):
We intentionally avoid using regexes to parse, as this tends to be a hot path executed pretty much on every request. Would it be possible the parsing functions to return an error in case the header value is not valid?

andriikushch:
@andrewslotin, I see your point. Please check the latest commit. What do you think?

andrewslotin (approved):

PR #286, opened 2021-12-09
Fix circleci build

andriikushch:
This is a quick fix for the build. "circleci" images are deprecated now, they are replaced with "cimg/go:[version]", but there is no support for older versions of go and they are not compatible with the current build script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,12 +23,15 @@ jobs:
         type: string
     docker:
       - image: circleci/golang:<< parameters.go_version >>
+        user: root
     working_directory: /go/src/github.com/instana/go-sensor
     steps:
       - checkout
       - restore_cache:
           keys:
             - go-mod-v4-{{ checksum "go.sum" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -23,12 +23,17 @@ jobs:
...
     working_directory: /go/src/github.com/instana/go-sensor
     steps:
       - checkout
       - restore_cache:
           keys:
             - go-mod-v4-{{ checksum "go.sum" }}
+      - run:
+          command: apt-get --allow-releaseinfo-change-suite update -y || true
+      - run:
+          command: apt-get update && apt-get install ca-certificates libgnutls30 -y || true
```
andrewslotin, line 36: Can we combine these steps into one `run` command?

andrewslotin (approved):

.circleci/config.yml
andriikushch: yes, I will change it

PR #287, opened 2021-12-09
Implement W3C tracestate limitations

andriikushch:
Implement https://www.w3.org/TR/trace-context/#tracestate-header

### DIFF
--- a/w3ctrace/state.go
+++ b/w3ctrace/state.go
@@ -11,24 +11,56 @@ import (
 // VendorInstana is the Instana vendor key in the `tracestate` list
 const VendorInstana = "in"
 
+// Max amount of KV pairs in `tracestate` header
+const maxKVPairs = 32
+
+// Length of entries that should be filtered first in case, if tracestate has more than `maxKVPairs` items
+const thresholdLen = 128
+
 // State is list of key=value pairs representing vendor-specific data in the trace context
 type State []string
 
-// ParseState parses the value of `tracestate` header. It strips any optional white-space chararacters
-// preceding or following the key=value pairs. Empty list items are omitted.
+// ParseState parses the value of `tracestate` header. Empty list items are omitted.
 func ParseState(traceStateValue string) (State, error) {
 	var state State
 
-	for _, st := range strings.SplitN(traceStateValue, ",", 32) {
-		st = strings.TrimSpace(st)
-		if st == "" {
-			continue
-		}
-
-		state = append(state, st)
+	entries := strings.Split(traceStateValue, ",")
+	if len(entries) == 1 && entries[0] == "" {
+		return state, nil
 	}
 
-	return state, nil
+	filteredEntries := filterEmptyItems(entries)
+	filteredEntriesLen := len(filteredEntries)
+
+	if filteredEntriesLen == 0 {
+		return state, nil
+	}
+
+	if filteredEntriesLen > maxKVPairs {
+		filtered := 0
+
+		for k, st := range filteredEntries {
+			if len(state) == maxKVPairs {
+				break
+			}
+
+			// check if enough elements were filtered already
+			if filteredEntriesLen-filtered <= maxKVPairs {
+				return append(state, filteredEntries[k:]...), nil
+			}
+
+			if len(st) > thresholdLen {
+				filtered++
+				continue
+			}
+
+			state = append(state, st)
+		}
+
+		return state, nil
+	}
+
+	return filteredEntries, nil
 }
 
 // Add returns a new state prepended with provided vendor-specific data. It removes any existing
@@ -108,4 +140,17 @@ func (st State) String() string {
 	buf.Truncate(buf.Len() - 1) // remove trailing comma
 
 	return buf.String()
+}
+
+func filterEmptyItems(entries []string) []string {
+	var entriesWithoutEmptyItems []string
+	for _, v := range entries {
+		if v == "" {
+			continue
+		}
+
+		entriesWithoutEmptyItems = append(entriesWithoutEmptyItems, v)
+	}
+
+	return entriesWithoutEmptyItems
 }
--- a/w3ctrace/state.go
+++ b/w3ctrace/state.go
@@ -22,45 +22,26 @@ type State []string
 
 // ParseState parses the value of `tracestate` header. Empty list items are omitted.
 func ParseState(traceStateValue string) (State, error) {
-	var state State
+	states := filterEmptyItems(strings.Split(traceStateValue, ","))
+	statesLen := len(states)
 
-	entries := strings.Split(traceStateValue, ",")
-	if len(entries) == 1 && entries[0] == "" {
-		return state, nil
+	if statesLen < maxKVPairs {
+		return states, nil
 	}
 
-	filteredEntries := filterEmptyItems(entries)
-	filteredEntriesLen := len(filteredEntries)
+	filteredStates := states[:0]
+	for stateNumber, st := range states {
+		unfilteredStatesAmount := statesLen - stateNumber
 
-	if filteredEntriesLen == 0 {
-		return state, nil
-	}
+		needToFilter := len(filteredStates)+unfilteredStatesAmount > maxKVPairs
 
-	if filteredEntriesLen > maxKVPairs {
-		filtered := 0
-
-		for k, st := range filteredEntries {
-			if len(state) == maxKVPairs {
-				break
-			}
-
-			// check if enough elements were filtered already
-			if filteredEntriesLen-filtered <= maxKVPairs {
-				return append(state, filteredEntries[k:]...), nil
-			}
-
-			if len(st) > thresholdLen {
-				filtered++
-				continue
-			}
-
-			state = append(state, st)
+		if len(st) > thresholdLen && needToFilter {
+			continue
 		}
-
-		return state, nil
+		filteredStates = append(filteredStates, st)
 	}
 
-	return filteredEntries, nil
+	return filteredStates[:min(len(filteredStates), maxKVPairs)], nil
 }
 
 // Add returns a new state prepended with provided vendor-specific data. It removes any existing
@@ -143,14 +124,20 @@ func (st State) String() string {
 }
 
 func filterEmptyItems(entries []string) []string {
-	var entriesWithoutEmptyItems []string
+	result := entries[:0]
 	for _, v := range entries {
-		if v == "" {
-			continue
+		if v != "" {
+			result = append(result, v)
 		}
-
-		entriesWithoutEmptyItems = append(entriesWithoutEmptyItems, v)
 	}
 
-	return entriesWithoutEmptyItems
+	return result
+}
+
+func min(a, b int) int {
+	if a > b {
+		return b
+	}
+
+	return a
 }
--- a/w3ctrace/state.go
+++ b/w3ctrace/state.go
@@ -23,16 +23,15 @@ type State []string
 // ParseState parses the value of `tracestate` header. Empty list items are omitted.
 func ParseState(traceStateValue string) (State, error) {
 	states := filterEmptyItems(strings.Split(traceStateValue, ","))
-	statesLen := len(states)
+	unfilteredStatesAmount := len(states)
 
-	if statesLen < maxKVPairs {
+	if unfilteredStatesAmount < maxKVPairs {
 		return states, nil
 	}
 
 	filteredStates := states[:0]
-	for stateNumber, st := range states {
-		unfilteredStatesAmount := statesLen - stateNumber
-
+	for _, st := range states {
+		unfilteredStatesAmount--
 		needToFilter := len(filteredStates)+unfilteredStatesAmount > maxKVPairs
 
 		if len(st) > thresholdLen && needToFilter {
--- a/w3ctrace/state.go
+++ b/w3ctrace/state.go
@@ -23,24 +23,30 @@ type State []string
 // ParseState parses the value of `tracestate` header. Empty list items are omitted.
 func ParseState(traceStateValue string) (State, error) {
 	states := filterEmptyItems(strings.Split(traceStateValue, ","))
-	unfilteredStatesAmount := len(states)
-
-	if unfilteredStatesAmount < maxKVPairs {
+	if len(states) < maxKVPairs {
 		return states, nil
 	}
 
+	itemsToFilter := len(states) - maxKVPairs
 	filteredStates := states[:0]
-	for _, st := range states {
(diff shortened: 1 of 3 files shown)


w3ctrace/state.go
```diff
@@ -19,13 +19,34 @@ type State []string
 func ParseState(traceStateValue string) (State, error) {
 	var state State
 
-	for _, st := range strings.SplitN(traceStateValue, ",", 32) {
-		st = strings.TrimSpace(st)
-		if st == "" {
-			continue
+	// comasLimit is used to prevent handling of malformed values that contain a lot of comas
+	const comasLimit = 256
+	rawEntries := strings.SplitN(traceStateValue, ",", comasLimit)
```
andrewslotin, line 24:
```suggestion
	const maxEntries = 256
	rawEntries := strings.SplitN(traceStateValue, ",", maxEntries)
```

w3ctrace/state.go
```diff
@@ -19,13 +19,34 @@ type State []string
 func ParseState(traceStateValue string) (State, error) {
 	var state State
 
-	for _, st := range strings.SplitN(traceStateValue, ",", 32) {
-		st = strings.TrimSpace(st)
-		if st == "" {
-			continue
+	// comasLimit is used to prevent handling of malformed values that contain a lot of comas
+	const comasLimit = 256
```
andrewslotin, line 23: Where does this number come from?

w3ctrace/state.go
```diff
@@ -19,13 +19,34 @@ type State []string
...
+
+	rawEntriesLen := len(rawEntries)
+	if rawEntriesLen > 32 {
+		filtered := 0
+		for _, st := range rawEntries {
+			if rawEntriesLen-filtered <= 32 || len(state) == 32 {
+				break
+			}
+
+			st = strings.TrimSpace(st)
```
andrewslotin, line 34: This would potentially modify list members now owned by Instana, which may break trace correlation downstream. I'd suggest we don't alter the value that we did not add to the state.

w3ctrace/state.go
```diff
@@ -19,13 +19,34 @@ type State []string
...
+			if len(st) > 128 || st == "" {
+				filtered++
+				continue
+			}
+
+			state = append(state, st)
+		}
+	} else {
+		for _, st := range rawEntries {
+			st = strings.TrimSpace(st)
```
andrewslotin, line 44: Same as above, this would alter a value that we don't own.

andrewslotin (changes requested):
I'm not entirely sure the truncation logic follows the spec. At the moment for the `traceparent` header containing more than 32 key-value pairs these changes:
* Starting from beginning any entry longer than 128 characters or empty, not counting leading/trailing spaces is removed

This approach does not guarantee that resulting set would contain 32 elements or less, e.g. if state does not contain any empty/large key-value pairs.

On the other hand the W3C prescribes the following algorithm:
1. The entries at the end of `tracestate` SHOULD be removed first
2. First step is to remove entries longer then 128 characters
3. If the previous step did not produce a list of 32 elements or less, truncate the last `len - 32` elements.

Let me know if I'm missing something

w3ctrace/state.go
andriikushch: This number is something just bigger than a reasonable amount of tracestate entries (in raw header). To prevent a situation when malicious user sent values like ",,,,,,,,,,,,,,,,,,,,,,,,,(and many comas after)" and system, will waste a CPU resource. If you think it is too much, we can use `Split` instead of `SplitN`. What do you think?

w3ctrace/state.go
andriikushch: This number is something just bigger than a reasonable amount of tracestate entries (in raw header). To prevent a situation when malicious user sent values like ",,,,,,,,,,,,,,,,,,,,,,,,,(and many comas after)" and system, will waste a CPU resource. If you think it is too much, we can use Split instead of SplitN. What do you think?

w3ctrace/state.go
andriikushch: I tried to keep the same behavior as before https://github.com/instana/go-sensor/blob/master/w3ctrace/state.go#L23.

w3ctrace/state.go
andriikushch: Answered in the previous comment.

w3ctrace/state.go
andrewslotin: Given that the size of HTTP headers section is limited, I'd say it's unlikely this will become a problem.

w3ctrace/state.go
andrewslotin: This does not look correct either 😅

andriikushch: :) Let's change it then.

andriikushch:
> I'm not entirely sure the truncation logic follows the spec. At the moment for the `traceparent` header containing more than 32 key-value pairs these changes:
> 
> * Starting from beginning any entry longer than 128 characters or empty, not counting leading/trailing spaces is removed
> 
> This approach does not guarantee that resulting set would contain 32 elements or less, e.g. if state does not contain any empty/large key-value pairs.

I think this case is covered here https://github.com/instana/go-sensor/pull/287/files#diff-4d31341d37f313670fd6f87a6a018dccc0301c44aab43258fe1f99d8c3341942R34, isn't it? 

> 
> On the other hand the W3C prescribes the following algorithm:
> 
> 1. The entries at the end of `tracestate` SHOULD be removed first

It looks like this is missed here: `The entries at the end of tracestate SHOULD be removed first`, but I can't find it in a spec. Can you please share a link to this line, because I found only two next statements?
> 2. First step is to remove entries longer then 128 characters
> 3. If the previous step did not produce a list of 32 elements or less, truncate the last `len - 32` elements.
> 
> Let me know if I'm missing something

w3ctrace/state.go
andriikushch: Removed

w3ctrace/state.go
andrewslotin: Yes, but according to the spec the `value` may include whitespace characters (`%x20`) and since they are indistinguishable from optional whitespace (`OWS`) in the end, I'd suggest we don't modify the elements that we don't add.

w3ctrace/state_test.go
```diff
@@ -30,6 +31,18 @@ func TestParseState(t *testing.T) {
 			Header:   "rojo=00f067aa0ba902b7,    ,,congo=t61rcWkgMzE",
 			Expected: w3ctrace.State{"rojo=00f067aa0ba902b7", "congo=t61rcWkgMzE"},
 		},
+		"with 33 list items": {
+			Header:   strings.TrimLeft(strings.Repeat("rojo=00f067aa0ba902b7,", 33), ","),
```
andrewslotin, line 35:
```suggestion
			Header:   strings.TrimRight(strings.Repeat("rojo=00f067aa0ba902b7,", 33), ","),
```
Otherwise it's leaving the trailing comma at the end, which is not an issue for the test, but contradicts the test name

w3ctrace/state_test.go
```diff
@@ -30,6 +31,18 @@ func TestParseState(t *testing.T) {
...
+			Expected: strings.Split(strings.TrimRight(strings.Repeat("rojo=00f067aa0ba902b7,", 32), ","), ","),
+		},
+		"with 34 list items, with long one at the beginning": {
+			Header:   "rojo=" + strings.Repeat("a", 129) + "," + strings.TrimLeft(strings.Repeat("rojo=00f067aa0ba902b7,", 33), ","),
```
andrewslotin, line 39:
```suggestion
			Header:   "rojo=" + strings.Repeat("a", 129) + "," + strings.TrimRight(strings.Repeat("rojo=00f067aa0ba902b7,", 33), ","),
```
Same as above

andrewslotin:
> I think this case is covered here https://github.com/instana/go-sensor/pull/287/files#diff-4d31341d37f313670fd6f87a6a018dccc0301c44aab43258fe1f99d8c3341942R34, isn't it?

Right, I was confused by the condition that breaks the loop 👍 

> It looks like this is missed here: "The entries at the end of tracestate SHOULD be removed first", but I can't find it in a spec. Can you please share a link to this line, because I found only two next statements?

Here is the quote: 

> In a situation where tracestate needs to be truncated due to size limitations, the vendor MUST truncate whole entries. Entries larger than 128 characters long SHOULD be removed first. Then entries SHOULD be removed starting from the end of `tracestate`. 

Same as above, I believe I misread the loop condition. I believe the part `rawEntriesLen-filtered <= 32` is redundant, at least removing it does not fail any of your tests.

w3ctrace/state_test.go
andriikushch: Good catch - typo 👍

w3ctrace/state.go
```diff
@@ -19,13 +19,29 @@ type State []string
...
+				continue
+			}
+
+			state = append(state, st)
+		}
+	} else {
+		for _, st := range rawEntries {
+			if st == "" {
+				continue
+			}
```
andrewslotin, line 42: As a suggestion, if we stop filtering empty list items for each request, we could spare some CPU cycles and use a guard-clause here instead of `else`

andriikushch: Like this https://github.com/instana/go-sensor/pull/287/files#diff-8e3fa4ec5050520f1e8157ac884ce463ca2553109301a7e640e2e0a5f9927aa7R38?

andriikushch:
> Same as above, I believe I misread the loop condition. I believe the part `rawEntriesLen-filtered <= 32` is redundant, at least removing it does not fail any of your tests.

I think you are right.

andriikushch:
@andrewslotin I changed the logic and added a few more tests. I hope now the code looks more clear and straightforward. Please take a look when you will have time.

w3ctrace/state.go
```diff
@@ -11,24 +11,56 @@ import (
...
 	var state State
 
-	for _, st := range strings.SplitN(traceStateValue, ",", 32) {
-		st = strings.TrimSpace(st)
-		if st == "" {
-			continue
+	entries := strings.Split(traceStateValue, ",")
+	if len(entries) == 1 && entries[0] == "" {
+		return state, nil
+	}
```
andrewslotin, line 30: This looks like a premature optimization: the `range` statement inside `filterEmptyItems` will not be executed for an empty list, so the check at L35 will suffice.

w3ctrace/state.go
```diff
@@ -11,24 +11,56 @@ import (
...
+
+	filteredEntries := filterEmptyItems(entries)
+	filteredEntriesLen := len(filteredEntries)
+
+	if filteredEntriesLen == 0 {
+		return state, nil
+	}
```
andrewslotin, line 37: Actually this check is also redundant, since the following `if` statement condition will result to `false` in case the list is empty.

w3ctrace/state.go
```diff
@@ -11,24 +11,56 @@ import (
...
+
+	if filteredEntriesLen > maxKVPairs {
```
andrewslotin, line 39: Let's change it to an early return to reduce the cognitive complexity.

w3ctrace/state.go
```diff
@@ -109,3 +141,16 @@ func (st State) String() string {
 
 	return buf.String()
 }
+
+func filterEmptyItems(entries []string) []string {
+	var entriesWithoutEmptyItems []string
```
andrewslotin, line 146: As a suggestion, you could consider preallocating the slice here, since its max capacity is known.

w3ctrace/state.go
```diff
@@ -11,24 +11,56 @@ import (
...
+
+		for k, st := range filteredEntries {
+			if len(state) == maxKVPairs {
+				break
+			}
+
+			// check if enough elements were filtered already
+			if filteredEntriesLen-filtered <= maxKVPairs {
+				return append(state, filteredEntries[k:]...), nil
+			}
```
andrewslotin, line 50: I still find it hard to understand the logic here, since one needs to consider the size two lists at once. I believe we can do this with one list, removing the items longer than 128 chars until either it contains 32 elements, or there are no large elements left. And then return first `min(len(state), 32)` elements of it?

andrewslotin (changes requested):

w3ctrace/state.go
andriikushch: This is changed now 👍

w3ctrace/state.go
andriikushch: This is changed now 👍

w3ctrace/state.go
andriikushch: This part now has a different implementation, without extra memory allocation.

w3ctrace/state.go
andriikushch: 👍

w3ctrace/state.go
andriikushch: Thank you for the idea. I have changed the implementation and added a new test case.

1. I removed extra memory allocations for filtering.
2. I changed the filtering function, now I think it looks close to what you propose.

Please check when you will have time.

w3ctrace/state.go
```diff
@@ -11,24 +11,40 @@ import (
...
 			continue
 		}
 
-		state = append(state, st)
+		filteredStates = append(filteredStates, states[i])
 	}
+	filteredStates = append(filteredStates, states[i:]...)
 
-	return state, nil
+	return filteredStates[:min(len(filteredStates), maxKVPairs)], nil
```
andrewslotin, line 47: Again just for readability reasons let's inline the `min()` function:
```suggestion
	if len(filteredStates) > maxKVPairs {
		filteredStates = filteredStates[:maxKVPairs]
	}
	
	return filteredStates
```

andrewslotin (approved):
Overall I think it looks good 👍 Just a couple of suggestions to improve readability even further.

w3ctrace/state.go
```diff
@@ -11,24 +11,40 @@ import (
...
 // State is list of key=value pairs representing vendor-specific data in the trace context
 type State []string
 
-// ParseState parses the value of `tracestate` header. It strips any optional white-space chararacters
-// preceding or following the key=value pairs. Empty list items are omitted.
+// ParseState parses the value of `tracestate` header. Empty list items are omitted.
 func ParseState(traceStateValue string) (State, error) {
-	var state State
+	states := filterEmptyItems(strings.Split(traceStateValue, ","))
+	statesLen := len(states)
```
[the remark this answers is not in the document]
andriikushch: done

w3ctrace/state.go
andriikushch: Added

PR #289, opened 2022-01-20
Add the usage of INSTANA_LOG_LEVEL environment variable

willianpc:


logger/logger.go
```diff
@@ -85,11 +86,29 @@ func New(printer Printer) *Logger {
...
+		"error": ErrorLevel,
+		"info":  InfoLevel,
+		"warn":  WarnLevel,
+		"debug": DebugLevel,
+	}
+
+	if value, ok := envVarLogLevel[strings.ToLower(instanaLogLevelEnvVar)]; ok {
+		l.SetLevel(value)
+	} else {
+		l.SetLevel(ErrorLevel)
```
andriikushch, line 108: @willianpc would it makes sense, to handle a case when value is provided but is invalid (write a message or something like this)? with the current implementation, we will silently set the level as an error, but it will be not visible to the user.

andriikushch (approved):
@willianpc looks good to me! 👍 Good job! There is a tiny question regarding handling the invalid input. Please consider it. Otherwise, let's merge it and release it.

logger/logger.go
willianpc: I will create a separate PR for this improvement.

PR #290, opened 2022-01-27
Smarter agent communication log messages

willianpc:


fsm.go
```diff
@@ -130,11 +134,13 @@ func (r *fsmS) lookupSuccess(host string) {
 func (r *fsmS) announceSensor(e *f.Event) {
 	cb := func(success bool, resp agentResponse) {
 		if !success {
-			r.agent.logger.Error("Cannot announce sensor. Scheduling retry.")
 			r.retriesLeft--
 			if r.retriesLeft == 0 {
+				r.agent.logger.Error("Couldn't announce sensor after reaching the maximum amount of attempts.")
```
andriikushch, line 139:
```suggestion
				r.agent.logger.Error("Couldn't announce the sensor after reaching the maximum amount of attempts.")
```

fsm.go
```diff
@@ -76,36 +76,40 @@ func (r *fsmS) scheduleRetryWithExponentialDelay(e *f.Event, cb func(e *f.Event)
...
 		if found {
 			r.lookupSuccess(host)
 			return
 		}
 
-		gateway, err := getDefaultGateway("/proc/net/route")
-		if err != nil {
-			r.agent.logger.Error("failed to fetch the default gateway, scheduling retry: ", err)
-			r.scheduleRetry(e, r.lookupAgentHost)
+		if runtime.GOOS == "Linux" {
```
andriikushch, line 85:
```suggestion
		if _, err := os.Stat("/proc/net/route"); err == nil {
```

What do you think about this? Shall we check if the file exists, instead of if OS is Linux?

andriikushch (approved):
Overall looks good to me. 👍 Please check my comments and feel free to merge.

PR #291, opened 2022-01-27
Enable use of "PreferSimpleProtocol" option for a "pgx" driver.

andriikushch:
After implementing QueryContext in the SQL instrumentation, instrumentation respects the`PreferSimpleProtocol` option for queries when used with `pgx` driver.


willianpc (approved):
LGTM.

PR #292, opened 2022-02-03
add PostgreSQLSpanType

andriikushch:
This PR adds a new span type `PostgreSQLSpanType`. Similar to what we have for another technology.


willianpc (approved):
LGTM

PR #293, opened 2022-02-08
Add pgx instrumentation

andriikushch:
This PR adds instrumentation for a pgx library.

Please see instrumentation/instapgx/README.md for details.


noctarius:
An official instrumentation ❤️

.circleci/config.yml
```diff
@@ -24,9 +24,25 @@ jobs:
...
+          POSTGRES_PASSWORD: mysecretpassword
     working_directory: /go/src/github.com/instana/go-sensor
     steps:
       - checkout
+      - run:
+          name: Waiting for Postgres to be ready
+          command: |
+            for i in `seq 1 10`;
+            do
+              nc -z localhost 5432 && echo Success && exit 0
```
willianpc, line 40: [Optional] Maybe we could add the env var `PGPORT` after line 31 to match the port 5432 from the "waiting" script. On the other hand, 5432 is the Postgres default port, so not adding it is also fine.

instrumentation/instapgx/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Usage
+-----
+
+```go
+// Create a sensor
+sensor := instana.NewSensor("pgx-sensort")
+
+// Parse config
+conf, err := pgx.ParseConfig("postgres://postgres:mysecretpassword@localhost/postgres?sslmode=disable")
```
willianpc, line 37: I don't think users have to use `sslmode=disable`, do they?
Maybe we can remove this piece from the string to avoid confusion?

instrumentation/instapgx/conn.go
```diff
@@ -0,0 +1,216 @@
...
+// Conn wraps *pgx.Conn and adds tracing to it.
+type Conn struct {
+	*pgx.Conn
+	sensor *instana.Sensor
+	config *pgx.ConnConfig
+}
+
+// Prepare wraps (*pgx.Conn).Prepare method and adds tracing to it.
+func (c *Conn) Prepare(ctx context.Context, name, sql string) (*pgconn.StatementDescription, error) {
+	childCtx, span := contextWithChildSpan(ctx, fmt.Sprintf("PREPARE %s AS %s", name, sql), c.config, c.sensor)
```
willianpc, line 36: Do we need to worry about sensitive data that needs to be redacted from `sql`?
For instance, if the contents of the `sql` variable are displayed in the UI or somewhere else, where values from a query could be visible.

instrumentation/instapgx/row.go
```diff
@@ -0,0 +1,39 @@
+// (c) Copyright IBM Corp. 2021
```
willianpc, line 1: New files no longer need the Instana header, and please fix the year to 2022 😵‍💫

willianpc (approved):
Overall, LGTM.
Please, take a look at my comments, though.
Cheers!

instrumentation/instapgx/README.md
andriikushch: done

instrumentation/instapgx/conn.go
andriikushch: We do not collect parameters for a statement. In case it is embedded into SQL string then it will be there.

PR #296, opened 2022-02-15
relax assert statement in the test

andriikushch:
When we calculate durations and timestamps for a span, we use the division operation with uint64. When we do this we lose precision. This leads to having an operation with a duration equal to 0, which can't be true. It is always a positive number, bigger than 0.

This PR relaxes the assert statement.


willianpc (approved):
LGTM

PR #297, opened 2022-02-15
fix test

andriikushch:
Makes tests more correct. Sleeping for some time does not provide any guarantee on how long this sleep indeed will be.


willianpc (approved):
LGTM

PR #302, opened 2022-03-21
Add RedisSpanType

willianpc:
This PR adds a new span type `RedisSpanType`, similar to what we have for other technologies.


andrewslotin (changes requested):
Overall LGTM. Could you add `redis` to the [`TestRegisteredSpanType_ExtractData`](https://github.com/instana/go-sensor/blob/5ffba36889df1a8419e3e029a36d89ff9057b421/registered_span_test.go#L16) test before we merge this?

andrewslotin (approved):

PR #303, opened 2022-03-21
Use go-proxy-pull-action/v1.0.3

andrewslotin:
The `go-proxy-pull-action` that triggers the update of go.dev documentation upon release uses `golang:alpine` image that has switched go `go1.18` recently. In this Go version the behavior of `go get` has been changed to work with modules only, which prevents the action from execution.

`go-proxy-pull-action@v1.0.3` has [added](https://github.com/andrewslotin/go-proxy-pull-action/pull/6) support for `go1.18` and should be used instead.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -14,4 +14,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Pull new module version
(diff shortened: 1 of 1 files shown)


willianpc (approved):
LGTM.
Thanks for the change.

PR #304, opened 2022-03-31
Adding RabbitMQSpanType

willianpc:
This PR adds a new span type RabbitMQSpanType, similar to what we have for other technologies.


andriikushch (approved):
LGTM 👍

PR #309, opened 2022-05-03
do not discard the Instana tracestate list member

basti1302:
Previously, when more than 32 list members were in the incoming
tracestate header value, and the Instana-specific list member (in=...)
was at position 33 or later, we would drop it together with the other
list members that need to be droppped. With this change, we give the
Instana list member higher priority than list members from other
vendors.

We also keep it separate in the TraceState struct.

Furthermore, this simplifies/reduces the API of the TraceState module.
We do not have a use case for adding/removing list members for arbitrary
vendors - we only need to propagate the list members we received and
make sure that the Instana trace state value is added in the left-most
position.

### DIFF
--- a/w3ctrace/state.go
+++ b/w3ctrace/state.go
@@ -5,6 +5,7 @@ package w3ctrace
 
 import (
 	"bytes"
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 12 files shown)


basti1302:
I tested this against the Instana tracer test suite and it is as 🟢 🟢 🟢 as the grass on the other side, including the formerly failing test case 35. 😏 

The tests (I believe integration tests?) are failing for go 1.10, but that looks unrelated to my changes, I think.

basti1302:
Pretty sure there are a bunch of non-idiomatic things in this PR, given my lack of proficiency with Golang. Just let me know what you want changed/fixed/...

w3ctrace/state.go
```diff
@@ -17,109 +18,116 @@ const maxKVPairs = 32
...
-	itemsToFilter := len(states) - maxKVPairs
-	filteredStates := states[:0]
+	// The value for the Instana-specific list member ("in=...").
+	instanaTraceStateValue string
+}
 
-	i := 0
-	for ; itemsToFilter > 0 && i < len(states); i++ {
+// makeState creates a new State with the given values
+func MakeState(listMembers []string, instanaTraceStateValue string) State {
```
willianpc, line 34: A common convention is to use `NewState`.
If the function doesn't need to be used outside of this package, you can start with a lower case letter, so its visibility stays limited to it.

The same applies to any other function. Eventually we have to make them visible outside of their package if we want to, for instance, use them in the tests, which are a different package name.

w3ctrace/state.go
```diff
@@ -17,109 +18,116 @@ const maxKVPairs = 32
...
 
-		if len(states[i]) > thresholdLen {
-			itemsToFilter--
-			continue
+// ParseState parses the value of `tracestate` header. Empty list items are omitted.
+func ParseState(traceStateValue string) (State, error) {
+	listMembers := filterEmptyItems(strings.Split(traceStateValue, ","))
+
+	// Look for the Instana list member first, before discarding any list members due to length restrictions.
+	instanaTraceStateValue := ""
```
willianpc, line 46: Optional, but mostly for you knowledge (if you don't know it already), you can declare a var without giving it a value, because in Go, every var has always its "zero" value by default. eg: numbers is 0, strings is "" and so on :) 

```go
var instanaTraceStateValue string
```

w3ctrace/state.go
basti1302: I'll rename it to NewState then 👍 

The function is not used in production code at all, but only in `state_test.go` and `context_test.go`. That's why it is currently public/upper case.

Any recommendations around naming or visibility for that scenario (a function that only needs to be visible in the tests)?

One alternative would be to make the two attributes of the struct exported, but my gut feeling is that this would be actually worse than exporting the NewState function.

w3ctrace/state.go
```diff
@@ -17,109 +18,116 @@ const maxKVPairs = 32
...
-	return newSt
-}
-
-// Fetch retrieves stored vendor-specific data for given vendor
-func (st State) Fetch(vendor string) (string, bool) {
-	i := st.Index(vendor)
-	if i < 0 {
-		return "", false
+	itemsToFilter := len(listMembers) - maxListMembers
+	filteredStates := listMembers[:0]
```
willianpc, line 73: `filteredStates` is an empty slice of strings, which is later replaced by `append`. So
In this case, I think it's safe to create a new one, no? So it should be safe to use `var filteredStates []string` or `filteredStates := []string{}`.

Also, a slice still references to its original array/slice, so a change in a slice of a slice/array will modify the original one as well. Which shouldn't happen here, as `filteredStates` doesn't reference any item in `listMembers`.

w3ctrace/state.go
```diff
@@ -17,109 +18,116 @@ const maxKVPairs = 32
...
 // String returns string representation of a trace state. The returned value is compatible with the
-// `tracestate` header format
+// `tracestate` header format. If the state has an Instana-specific list member, that one is always rendered first. This
+// is optimized for the use case of injecting the string represenation of the tracestate header into downstream
+// requests.
 func (st State) String() string {
-	if len(st) == 0 {
+	if len(st.listMembers) == 0 && st.instanaTraceStateValue == "" {
 		return ""
+	} else if len(st.listMembers) == 0 {
```
willianpc, line 119: A common practice of the Go Team is to use guard clauses. 
So instead of using `else if`, could you please start a brand new if condition?

w3ctrace/state.go
basti1302: For the record, this code was exactly in this shape and form before my changes: https://github.com/instana/go-sensor/blob/master/w3ctrace/state.go#L30-L47

Basically the only change here is that calculate maxListMembers beforehand (either as maxKVPairs or maxKVPairs-1) and use that instead of maxKVPairs. Plus, I renamed `states` to `listMembers`.

But I'm fine with improving the code as part of this PR. I am not sure if I understand your suggestion, though.

So, the initial `listMembers[:0]` clears out the existing array. Maybe Andrey chose this over `[]string{}` for performance reasons? Quick googling suggests that `listMembers[:0]` will keep the already allocated memory and our expectation is probably that in most cases, we will end up with the same array as before anyway. (Kicking out a list member for being longer than 128 characters should be rare.)

WDYT?

w3ctrace/state.go
willianpc: As far as I can see in the code base, it is ok to make it publicly visible in order to be used in the tests, as the package name is different.
Maybe @andriikushch can have better ideas for cases like these.

w3ctrace/state.go
basti1302: Done.

w3ctrace/state.go
basti1302: Done.

w3ctrace/state.go
basti1302: I additionally renamed filteredStates to filteredListMembers now, should have done that in the first place.

basti1302: About this:
> Also, a slice still references to its original array/slice, so a change in a slice of a slice/array will modify the original one as well. Which shouldn't happen here, as filteredStates doesn't reference any item in listMembers.

AFAICT, the listMembers is created locally in the function ParseState and discarded after we are done with it, we only return the filteredListMembers. So my thinking is that it shouldn't make a difference whether we make destructive changes to the original listMembers array via filteredStates (now renamed to filteredListMembers) or not.

willianpc: Good point. I didn't notice that it was like that before. You are correct. whatever is written in the second array will be passed to the old one, overwriting its content from the beginning.
Nevermind what I said. the array[:0] trick must stay there.

basti1302: > I didn't notice that it was like that before.

Yeah, it is quite impossible to see that from the diff :-/

andriikushch:
Hi, @basti1302. Great job 👍 Thank you very much for finding time to fix it.

PR looks good to me. I have a few tiny suggestions and I did put them on the separate PR, please take a look when you will have time https://github.com/instana/go-sensor/pull/312.

basti1302:
@andriikushch I cherry-picked the relevant changes from the other branch into this one and closed the other PR. I hope that is okay with you.

andriikushch (approved):
Looks good to me! Great job! 👍

PR #310, opened 2022-05-04
Added Kafka headers as string option

willianpc:
This PR adds the Kafka header migration to the [sarama](https://pkg.go.dev/github.com/Shopify/sarama) instrumentation.
Through the `INSTANA_KAFKA_HEADER_FORMAT` environment variable, customers are able to choose between legacy `binary` headers, the new `string` headers or `both` at the same time.


instrumentation/instasarama/instasarama.go
```diff
@@ -4,3 +4,55 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+const KafkaHeaderEnvVarKey = "INSTANA_KAFKA_HEADER_FORMAT"
+
+type KafkaHeaderType int
```
andriikushch, line 16: What do you think if instead of using int here, we will use string type? 


This will allow us to avoid adding extra type and conversion logic. All the validations and default values handling can be implemented inside "GetKafkaHeaderFormat".

instrumentation/instasarama/instasarama_test.go
```diff
@@ -50,3 +50,13 @@ func extractAgentSpan(span interface{}) (agentSpan, error) {
 
 	return data, nil
 }
+
+// func TestRunExample(t *testing.T) {
```
andriikushch, line 54: We can think about extracting it to a separate file or examples folder.

instrumentation/instasarama/propagation.go
```diff
@@ -40,47 +55,75 @@ type ProducerMessageCarrier struct {
...
+					copy(traceContext[:16], existingC[:16])
+				}
 			}
+
+			c.addOrReplaceHeader(fieldCKey, traceContext)
 		}
 
-		c.addOrReplaceHeader(fieldCKey, traceContext)
+		if kafkaHeaderFormat == BOTH || kafkaHeaderFormat == STRING {
+			c.addOrReplaceHeader(fieldSKey, []byte(val))
```
andriikushch, line 105: For the `BINARY` case we check the length of the value, do we want to do the same here?

andriikushch, line 105: For the `BINARY` case we check the length of the value, do we want to do the same here?

andriikushch (changes requested):
Looks good to me! 👍  There are a few tiny questions about details, nothing critical. And I also want to ask you what do you think about removing conditional logic from the tests and copying these parts to the separate tests?

andriikushch (approved):
LGTM 👍

PR #311, opened 2022-05-05
fix: retry agent host lookup when /proc/net/route does not exist

basti1302:
I noticed when testing on MacOS that the Go tracer would give up connecting (or reconnecting) to the agent after one failed attempt to find the agent. There seems to be a missing `scheduleRetry` for the case when `/proc/net/route` does not exist. With the added scheduleRetry call, the trace correctly attempts to find the agent every 30 seconds.


fsm.go
```diff
@@ -110,6 +110,9 @@ func (r *fsmS) lookupAgentHost(e *f.Event) {
 				r.scheduleRetry(e, r.lookupAgentHost)
 			})
 		}
+
+		r.agent.logger.Info("/proc/net/route does not exist. Scheduling retry.")
```
andriikushch, line 114: Good catch @basti1302 👍 

Just a small thing here: I think we need to wrap it with `else` otherwise for procfs envs it will schedule retry twice. 
And what do you think if we change it from info to debug?

andriikushch (changes requested):

fsm.go
basti1302: Hm, I guess you are right about the `else`. I was under the impression that all other branches ended with a `return`, but there is indeed no `return` after the `go r.checkHost(gateway...` call. I'll add an `else` 👍 

I don't have a very strong opinion about the log level, but all other retries are announced at level `error`, so my thinking was that `info` is already quite a bit lower in severity. WDYT?

Actually, maybe I should also add the phrase `Cannot connect to the agent.` to the message. The non-existence of `/proc/net/route` is not that noteworthy but the fact that we could not yet connect to the agent is.

How about something like

```
r.agent.logger.Info("Cannot connect to the agent. Scheduling retry. (Connecting through the default gateway has not been attempted because proc/net/route does not exist.)")
```

If you rather had another level (debug, warn, error) just let me know.

basti1302: I've split the message into two messages now, one at level error in line with the other "Scheduling retry" messages and one at debug level. LMK if that works for you, or what you'd rather see there.

andriikushch: Looks good to me! 👍

basti1302: Good to know. FWIW, your review status still says "requested changes" so merging is blocked.

PR #312, opened 2022-05-09
Small update to the w3c headers handling

andriikushch:
@basti1302, @willianpc  Thank you a lot for the original PR. 👍 Great job. 👏 

I think the original PR is good and do its job well. In this PR I want to propose some ideas/changes to it. Please check if they make sense for you:

1. There is an update for golang.org/x/sys to mitigate this issue https://github.com/golang/go/issues/49219.
2. I removed a duplicated constant for the "32" value in state.go file (same for the test).
3. Removed the second returned parameter from the "ParseState" method which is always nil.
4. Extracted the "searchInstanaHeader" function from the "ParseState" method to make it easier to read.
5. Replaced "SetInstanaTraceStateValue" method with a function "FormStateWithInstanaTraceStateValue" (not sure if the name is good enough). I did it because this function/method is doing more than setting the value. It changes the list of the members and acts more like a constructor, that returns a new object from the new data and data from the old state.
6. Removed the "NewState" function, because it was used only in the tests.

### DIFF
--- a/w3ctrace/state_test.go
+++ b/w3ctrace/state_test.go
@@ -103,39 +103,38 @@ func TestParseState(t *testing.T) {
 
 	for name, example := range examples {
 		t.Run(name, func(t *testing.T) {
-			st, err := w3ctrace.ParseState(example.Header)
-			require.NoError(t, err)
+			st := w3ctrace.ParseState(example.Header)
 			assert.Equal(t, example.Expected, st)
 		})
 	}
 }
 
-func TestState_SetInstanaTraceStateValueIntoEmptyTraceState(t *testing.T) {
+func TestState_FormStateWithInstanaTraceStateValueIntoEmptyTraceState(t *testing.T) {
 	st := w3ctrace.NewState([]string{}, "")
-	st = st.SetInstanaTraceStateValue("fa2375d711a4ca0f;02468acefdb97531")
+	st = w3ctrace.FormStateWithInstanaTraceStateValue(st, "fa2375d711a4ca0f;02468acefdb97531")
 	require.Equal(t, w3ctrace.NewState([]string{}, "fa2375d711a4ca0f;02468acefdb97531"), st)
 }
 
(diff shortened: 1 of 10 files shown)


w3ctrace/state_test.go
```diff
@@ -13,151 +13,144 @@ import (
...
 	examples := map[string]struct {
 		Header   string
 		Expected w3ctrace.State
 	}{
 		"empty": {
-			Expected: w3ctrace.NewState([]string{}, ""),
+			Expected: w3ctrace.ParseState(""),
 		},
 		"single tracing system": {
 			Header:   "rojo=00f067aa0ba902b7",
```
basti1302, line 26: I understand the intent of removing `NewState` because it is "only" used in tests, however, I think the whole TestParseState test does not make a lot of sense anymore with this change. 

It effectively compares the result of ParseState with another result of ParseState for the same string. In 
https://github.com/instana/go-sensor/pull/312/files#diff-4d31341d37f313670fd6f87a6a018dccc0301c44aab43258fe1f99d8c3341942R100 we create the **actual** value as `ParseState(example.Header)`. Now with this change we also create the **expected** value basically in the same way. We are effectively testing that we use the same string in example.Header as we use in example.Expected (and that ParseState does not return an error).

I think to actually test `ParseState` we need a way to create a `State` struct without relying on `ParseState`.

basti1302 (commented):
This all looks good, I think, and definitely improves the PR, thanks!

The only change I disagree about removing the `NewState` function though, because that kinda makes the test for ParseState useless, I think. See inline comment for explanation.

w3ctrace/state_test.go
andriikushch: I agree. 👍  I will revert this change.

andriikushch:
@basti1302, I did think about the "NewState" and I agree that it makes sense that we have it. 👍

I reverted one commit.

basti1302 (approved):
👍

basti1302:
I have cherry-picked the first two commits into the other branch.

PR #313, opened 2022-05-11
Improve performance for FormatID() function

andriikushch:
This PR changes is how the value converts from int64 to uint64. It gives some performance benefits.

`BenchmarkFormatID-16                     4194214               264.2 ns/op           176 B/op          8 allocs/op
`
->
`BenchmarkFormatID-16                    10477688                98.68 ns/op           32 B/op          2 allocs/op
`

### DIFF
--- a/util_internal_test.go
+++ b/util_internal_test.go
@@ -181,3 +181,9 @@ eth0	00000000	010011AC	0003	0	0	0	00000000	0	0	0
 		}()
 	}
 }
+
+func BenchmarkFormatID(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		FormatID(int64(i))
+	}
+}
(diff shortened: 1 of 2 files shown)


util_internal_test.go
```diff
@@ -181,3 +181,9 @@ eth0	00000000	010011AC	0003	0	0	0	00000000	0	0	0
 		}()
 	}
 }
+
+func BenchmarkFormatID(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		FormatID(int64(i))
+	}
+}
```
willianpc, line 189: Would it be relevant to add a test the compares the result between the old (current) implementation with the new one? Or this is pretty much already covered in the other tests?

willianpc (approved):
LGTM. Nice improvement!
I only have a small question, but that doesn't prevent me to approve the PR :)

util_internal_test.go
andriikushch: There are tests that test this functionality and they still work.

PR #314, opened 2022-05-16
go-redis improved tests: mocked redis replaced by mocked go-redis API

willianpc:
This PR replaces the mocked Redis server - which was too complex and hard to maintain - by a mock around go-redis API instead.
This refactoring caused a small change in the wrappers signature, there is no incompatibility to the current usage of the instrumentation.


andriikushch (approved):
LGTM 👍

PR #315, opened 2022-05-24
Readding instasarama.go

willianpc:
Since this file only holds the package, it seems to have been deleted by VS Code automatically without us noticing.


willianpc:
The file was removed by @andriikushch in one of the commits.
We agreed that an empty file is not needed, and the information that it brings in pkg.go.dev is already available via README.md.

PR #317, opened 2022-06-01
make recorder configurable and add option to queue traces when not ready

mier85:
when our services start up, it will take a while till the instana sensor is ready to handle our traffic. however, we already make calls during that time to fetch configurations, roles, etc. 

we do not want to be blind for these calls and so we want to be able to queue calls when the sensor is not yet ready. 

this is a none-breaking change which does just that. 

Implementing my own recorder is unfortunately not an option as the interface contains an unexported struct.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -26,13 +26,40 @@ type Recorder struct {
 	sync.RWMutex
 	spans    []Span
 	testMode bool
+
+	queueWhenNotReady bool
+	flushInterval     time.Duration
+}
+
+// RecorderOption lets you specify more options for the recorder
+type RecorderOption func(recorder *Recorder)
+
+// RecorderWithQueueWhenNotReady will enable queuing even when the sensor is not yet ready
+func RecorderWithQueueWhenNotReady() RecorderOption {
+	return func(recorder *Recorder) {
+		recorder.queueWhenNotReady = true
+	}
+}
+
+// RecorderWithFlushInterval lets you specify how often traces should be flushed
+func RecorderWithFlushInterval(flushInterval time.Duration) RecorderOption {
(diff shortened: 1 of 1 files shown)


andriikushch:
Hi @mier85, 

Thank you very much for your interest and this PR.

This new configuration option allows sending of the data before the tracing configuration is received from the agent. This can lead to tracer misconfiguration and/or sensitive data leakage.

mier85:
Hi @andriikushch , Can you help me explain how it leads to leaking data exactly? 
If I start the applicaiton and wait till instana sensor is ready before I do anything else, will the problem of sending sensitive data will disappear? 

How would the tracer be misconfigured in this scenario? Except for the queued data, everything else seems to go the same? 
In my tests with this fork, it lead to good results.

willianpc:
> Hi @andriikushch , Can you help me explain how it leads to leaking data exactly? If I start the applicaiton and wait till instana sensor is ready before I do anything else, will the problem of sending sensitive data will disappear?
> 
> How would the tracer be misconfigured in this scenario? Except for the queued data, everything else seems to go the same? In my tests with this fork, it lead to good results.

Hi @mier85 
There are a set of configurations received by the tracer once the communication is established with the agent.
Among them, some security configs, including the removal/obfuscation of sensitive data. 
If spans are collected prior to the tracer receiving these configs from the agent, such configs will be ignored which may cause harm, as sensitive data may be exposed.

As this discussion is out of the scope of this PR, I suggest that you folks file a Zendesk ticket.

I hope this clarifies your question.

Best,
- Willian

willianpc:
Closing the PR as there was no further interaction from the author.

PR #319, opened 2022-06-08
Add wrappers for the AWSSDK session's constructors

andriikushch:
This PR adds new wrappers to the AWSSDK instrumentation package. It does not extend existing functionality, it is just an alternative method to get an instrumented session.


willianpc (approved):
LGTM

PR #320, opened 2022-06-09
Returned struct is now public visible

willianpc:
This change essentially allows customers to declare variables in addition to type inference assignments.
The type returned by the instrumentation is now public, so customers can write something like this:

```go
var router *instahttprouter.WrappedRouter
```

That was not possible previously as `wrappedRouter` was private.


andriikushch:
LGTM! Good job! 👍

andriikushch (approved):

PR #323, opened 2022-06-20
Fix sarama instrumentation for the newest library version

andriikushch:
Fix decoration for the `sarama.ConsumerGroup`. Now wrapper wraps only selected methods and implicitly implements others via embedded type.


willianpc (approved):
LGTM.
I am a big fan of this struct  embedding

PR #326, opened 2022-06-29
Combine "SQLOpen" and "InstrumentSQLDriver" in one call

andriikushch:
This PR improves the instrumentation of the `sql` package. Instrumentation can be done now with only one method call. Before it was required to do two separate calls `SQLOpen` and `InstrumentSQLDriver`. Also, it was necessary to know the type of the exact driver and provide it to the `InstrumentSQLDriver`, now it will be done automatically by reusing the driver from the registry.


instrumentation_sql.go
```diff
@@ -51,13 +53,26 @@ func InstrumentSQLDriver(sensor *Sensor, name string, driver driver.Driver) {
...
 	}
 
 	return sql.Open(driverName, dataSourceName)
 }
 
+//go:linkname drivers database/sql.drivers
+var drivers map[string]driver.Driver
+
+// SQLInstrumentAndOpen combines `InstrumentSQLDriver` and `SQLOpen` calls
+func SQLInstrumentAndOpen(sensor *Sensor, driverName, dataSourceName string) (*sql.DB, error) {
```
willianpc, line 68: Should we update the `http-database-greeter` example with the new code, or create a new one?
As per our agreement of having an `example` folder for all new instrumentations, this would be my preferable choice. 
As the SQL instrumentation is part of the core and we already have an `example` folder anyway, it would be just a matter of deciding if you want to change http-database-greeter or create a new one.

The only additional thing I would like to suggest is to add a docker-compose file into this directory that - initially - starts Postgres as a convenience.

What do you think?

willianpc (approved):
LGTM!

The code looks fine. As per our previous talk, we just need to keep an eye on the `go:linkname` directive in case the Go team adds some limitation to it in the future.

I added some comments, all about documentation and code examples for your consideration.
Cheers!

instrumentation_sql.go
```diff
@@ -51,13 +53,26 @@ func InstrumentSQLDriver(sensor *Sensor, name string, driver driver.Driver) {
...
 		driverName += "_with_instana"
 	}
 
 	return sql.Open(driverName, dataSourceName)
 }
 
+//go:linkname drivers database/sql.drivers
+var drivers map[string]driver.Driver
+
+// SQLInstrumentAndOpen combines `InstrumentSQLDriver` and `SQLOpen` calls
```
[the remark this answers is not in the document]
andriikushch: I think adding some more information to the description makes sense 👍. Because, as you mentioned, this is not exactly the same as using `InstrumentSQLDriver ` and `SQLOpen`, here we reuse the "original" driver.

instrumentation_sql.go
andriikushch: Makes sense. I updated an example, please check. It requires one more commit, after the main module will be released.

PR #332, opened 2022-07-27
Limit the size of the payload that can be sent to the agent

andriikushch:
This PR prevents sending span batches to the agent if they exceed more than 5MB. In this case, a batch is dropped and a warning message is printed.

Implementation is inspired by https://github.com/instana/nodejs/blob/main/packages/collector/src/agentConnection.js#L33

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -18,6 +19,8 @@ import (
 	"github.com/instana/go-sensor/acceptor"
 	"github.com/instana/go-sensor/autoprofile"
 )
+
+var payloadTooLargeErr = errors.New(`request payload is too large`)
 
 const (
 	agentDiscoveryURL = "/com.instana.plugin.golang.discovery"
@@ -35,6 +38,8 @@ const (
 
 	announceTimeout = 15 * time.Second
 	clientTimeout   = 5 * time.Second
+
+	maxContentLength = 1024 * 1024 * 5
 )
 
 type agentResponse struct {
@@ -80,6 +85,10 @@ func newServerlessAgentFromS(entityID, provider string) *fromS {
 	}
 }
 
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 type agentS struct {
 	from *fromS
 	host string
@@ -88,7 +97,7 @@ type agentS struct {
 	mu  sync.RWMutex
 	fsm *fsmS
 
-	client          *http.Client
+	client          httpClient
 	snapshot        *SnapshotCollector
 	logger          LeveledLogger
 	clientTimeout   time.Duration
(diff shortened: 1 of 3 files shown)


agent.go
```diff
@@ -261,7 +276,14 @@ func (agent *agentS) fullRequestResponse(ctx context.Context, url string, method
 
 	if err == nil {
 		if j != nil {
-			req, err = http.NewRequest(method, url, bytes.NewBuffer(j))
+			b := bytes.NewBuffer(j)
+			if b.Len() > maxContentLength {
+				sensor.logger.Warn(`A batch of spans have been rejected because they are too large to be sent to the agent.`)
+
```
sanojsubran, line 282: Suggestion: Does it make sense to capture the size in the log here?

sanojsubran:
Looks good to me !

sanojsubran (approved):

agent.go
```diff
@@ -261,7 +276,14 @@ func (agent *agentS) fullRequestResponse(ctx context.Context, url string, method
 
 	if err == nil {
 		if j != nil {
-			req, err = http.NewRequest(method, url, bytes.NewBuffer(j))
+			b := bytes.NewBuffer(j)
+			if b.Len() > maxContentLength {
+				sensor.logger.Warn(`A batch of spans have been rejected because they are too large to be sent to the agent.`)
```
basti1302, line 281: *nitpick:* The verb refers to _the batch_, which is singular. 🤓  

```suggestion
				sensor.logger.Warn(`A batch of spans has been rejected because it is too large to be sent to the agent.`)
```

basti1302, line 281: *remark:* In Node we go the extra mile and inspect the batch of spans to check for the largest span in the batch, and then logging it in some abbreviated form IIRC. This is very helpful for troubleshooting situations where this error occurs (often it is caused by a few spans with very large annotations). But AFAICT we don't have this option here because the data is already serialized/marshalled? Or does `data` contain the spans in some kind of structure that can still be inspected? Just curious.

basti1302 (approved):
LGTM

agent.go
andriikushch: Nice catch 👍 Thank you!

andriikushch: Hi @basti1302, 

Thank you for bringing up this topic. I have seen these log messages for the node tracer. I was not sure about how useful it will be here. Because to find the top 5 biggest serialized spans we have to serialize them one by one and then use heap/sort/something else. All of these might potentially consume a big amount of memory and create extra load on the CPU and log system because we do not know how huge is data, we now only that it is > 5MB. This is a reason why I did not put it there initially.

Nevertheless, if it helps to solve problems with debugging, I think it is ok to have it. Anyway, we serialized this data anyway before sending it, so in the worst case, it should more than ~doubled the memory we already use. Also as we log it only once and with a warning log level, it should not create unexpecting long-lasting issues.

I added this functionality (https://github.com/instana/go-sensor/pull/332/commits/68cd5db652a43abec696d8c70d414117a63a9312), please review it. 

Thank you in advance.

agent.go
```diff
@@ -173,8 +186,18 @@ func (agent *agentS) SendSpans(spans []Span) error {
 
 	_, err := agent.request(agent.makeURL(agentTracesURL), "POST", spans)
 	if err != nil {
-		agent.logger.Error("failed to send spans to the host agent: ", err)
-		agent.reset()
+		if err == payloadTooLargeErr {
+			agent.printPayloadTooLargeErrInfoOnce.Do(
+				func() {
+					agent.logDetailedInformationAboutDroppedSpans(5, spans, err)
```
sanojsubran, line 192: Query: What is the significance of 5 here as first argument?

andriikushch: Good question: this is a number of spans we want to log in this case. You are right is a kind of "magic constant". Maybe it is a good idea to extract it and give it a name.

sanojsubran (approved):
Looks good to me !

basti1302 (approved):
👍

agent.go
basti1302: Awesome! 👍

PR #334, opened 2022-08-16
Simplify code

andriikushch:
This PR contains simplification to the state machine used in the code.

1. It makes `fsmS` testable by introducing `fsmAgent` interface, outgoing connection can be mocked.
2. FSM now uses its own logger.
3. Retries functions/delays are configurable (can be injected), which simplifies testing a lot.
4. Callbacks are removed from the `fsmS`.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -287,7 +287,7 @@ func (agent *agentS) fullRequestResponse(ctx context.Context, url string, method
 		if j != nil {
 			b := bytes.NewBuffer(j)
 			if b.Len() > maxContentLength {
-				sensor.logger.Warn(`A batch of spans has been rejected because it is too large to be sent to the agent.`)
+				agent.logger.Warn(`A batch of spans has been rejected because it is too large to be sent to the agent.`)
 
 				return "", payloadTooLargeErr
 			}
--- a/agent.go
+++ b/agent.go
@@ -360,6 +360,14 @@ func (agent *agentS) setHost(host string) {
 	agent.host = host
 }
 
+func (agent *agentS) log() LeveledLogger {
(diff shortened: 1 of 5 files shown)


fsm.go
```diff
@@ -28,20 +28,38 @@ const (
...
+func newFSM(agent fsmAgent, logger LeveledLogger) *fsmS {
+	logger.Warn("Stan is on the scene. Starting Instana instrumentation.")
+	logger.Debug("initializing fsm")
 
 	ret := &fsmS{
-		agent:       agent,
-		retriesLeft: maximumRetries,
+		agent:                      agent,
+		retriesLeft:                maximumRetries,
+		expDelayFunc:               expDelay,
```
sanojsubran, line 60: Query : Here I was looking from where `expDelayFunc` was taking value. Does it make sense to move the `expDelay` to the top of the file. Or is it preferred to the keep the constructor function at the top?

agent.go
```diff
@@ -360,6 +360,14 @@ func (agent *agentS) setHost(host string) {
 	agent.host = host
 }
 
+func (agent *agentS) log() LeveledLogger {
```
sanojsubran: Minor: Inconsistency in the names of the getter functions

willianpc (approved):
LGTM.
I don't have further comments.

agent.go
andriikushch: I think this function was already removed in the following commits, wasn't it?

fsm.go
andriikushch: `expDelay` is used in multiple places, it doesn't belong to this particular type. It is also used in other places.

PR #339, opened 2022-08-31
Added fix for setting the priority for custom headers

sanojsubran:
This PR contains the fix for giving preference to the user set custom headers over the ones coming from the agent. So with this change the preference of custom headers will be in the following order: agent < developer set < env values


andriikushch (approved):
Looks good to me 👍 Good job!

willianpc (approved):
LGTM

PR #341, opened 2022-09-09
feat(kafka): use kafka header format 'both' by default

basti1302:
This starts phase 1 of the Kafka header migration.

See https://www.ibm.com/docs/en/instana-observability/current?topic=references-kafka-header-migration for more information.

# Timing

**This should go live for customers in early October, but not before October 1.** 

This PR is currently set to draft so we do not merge and release it too early. I would appreciate a review, though.


instrumentation/instasarama/consumer_group_handler_test.go
```diff
@@ -23,6 +23,15 @@ func TestConsumerGroupHandler_ConsumeClaim(t *testing.T) {
 		{
 			Topic: "topic-1",
 			Headers: []*sarama.RecordHeader{
+				{
```
basti1302, line 26: I'm entirely not sure about any of the changes in the tests for the *consumers*. Apparently we add headers to messages on the consumer side when no headers have been present in the original incoming message? That concept was new to me and a bit suprising. But from what I understand it is used so that the customers' applications can read the headers and continue the trace from the trace ID/span ID in the headers, is that correct?

If that understanding is correct, the header names and their contents are actually part of our public API in go and we need to make customers aware of this change, so they can update their application code somewhere between October 2022 and October 2023.

instrumentation/instasarama/consumer_group_handler_test.go
```diff
@@ -102,6 +123,18 @@ func TestConsumerGroupHandler_ConsumeClaim(t *testing.T) {
...
+			Value: []byte(span.TraceID),
+		})
+		assert.Contains(t, h.Messages[1].Headers, &sarama.RecordHeader{
+			Key:   []byte("X_INSTANA_S"),
+			Value: []byte(span.SpanID),
+		})
+		assert.Contains(t, h.Messages[1].Headers, &sarama.RecordHeader{
+			Key:   []byte("X_INSTANA_L_S"),
+			Value: []byte("1"),
+		})
```
basti1302, line 137: I'm entirely not sure about any of the changes in the tests for the *consumers*, in particular the cases where the original message has no headers and then we check in the test that headers have been added _at the consumer side_? That is different from what our other tracers do, which only add headers at the producer side, so this concept was new to me and a bit suprising. But from what I understand it is done so that the customers' applications can read the headers and continue the trace from the trace ID/span ID in the headers, is that correct?

If that understanding is correct, the header names and their contents are actually part of our public API in go and we need to make customers aware of this change, so they can update their application code somewhere between October 2022 and October 2023.

instrumentation/instasarama/consumer_group_handler_test.go
```diff
@@ -24,13 +24,24 @@ func TestConsumerGroupHandler_ConsumeClaim(t *testing.T) {
...
+					Value: []byte("0000000000000000000000000abcde12"),
+				},
+				{
+					Key:   []byte("x_instana_s"),
+					Value: []byte("00000000deadbeef"),
+				},
+				{Key: []byte("x_instana_l_s"), Value: []byte("1")},
+				{
+					// We deliberately send a different trace and span id in the binary header as in the string header to validate
+					// that the string headers get preference when both formats are present in the incoming message.
```
basti1302, line 37: Not sure if this is a valid way to test this, please double check.

willianpc: Looks fine for me.

willianpc (commented):
LGTM. Waiting for a review from the others.

instrumentation/instasarama/README.md
```diff
@@ -134,12 +135,14 @@ In the future, the binary headers will be discontinued and only the headers in t
 To choose a header format provide the `INSTANA_KAFKA_HEADER_FORMAT` environment variable to the application.
 The following are valid values:
 
-* `binary`: Will keep handling Kafka headers in the binary format. 
-* `string`: Will send Kafka header in the string format
-* `both`: Will set Instana headers n both binary and string formats. 
+* `binary`: Producers will only add binary headers to Kafka to messages.
+* `string`: Producers will only add string headers to Kafka to messages.
+* `both`: Producers will add both sets of headers to Kafka to messages.
```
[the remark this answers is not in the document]
basti1302: Thanks, I meant to write "will add ... to Kafka messages". The second "to" should not be there, I'll fix that.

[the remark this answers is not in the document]
basti1302: fixed

instrumentation/instasarama/README.md
```diff
@@ -123,8 +123,9 @@ in the message handler to continue the trace.
 
 ### Working With Kafka Header Formats
 
-Starting at v1.2.0, the instrumentation can handle both Kafka headers as binary (legacy) or as string.
-By default, and for versions prior to v1.2.0 , the default behavior is to process Kafka headers in the binary format.
+Starting at v1.2.0, the instrumentation can handle both Instana's trace correlation headers in the binary (legacy) format or in the newer string format.
```
sanojsubran, line 126: `Suggestion`: Since v1.2.0, the instrumentation supports Instana's trace correlation headers in both binary (legacy) and string (new) formats.

instrumentation/instasarama/async_producer_test.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 func TestAsyncProducer_Input(t *testing.T) {
-	headerFormats := []string{"binary", "string", "both"}
```
sanojsubran, line 20: `nitpicking` `opinionated` : `text` or `plaintext` format would have been better than `string` format. Just to back up my unnecessary argument, [binary strings](https://www.ibm.com/docs/en/db2/11.5?topic=list-binary-strings) are a thing in the IBM universe. ;)

instrumentation/instasarama/README.md
basti1302: Thanks, updated.

basti1302:
@willianpc @sanojsubran I've applied your readme suggestions, please re-review or approve if there are no further comments.

willianpc (approved):
LGTM

sanojsubran (approved):
LGTM

basti1302:
@instana/eng-go This can now be merged and released some time next week.

Issue #340, opened 2022-09-08
[Bug]:  ContextWithSpan / SpanFromContext not interoperable with opentracing


martinei:
### Problem Description

instana.ContextWithSpan (and SpanFromContext) use their own context-key to store the current span in context.

As a result http handlers wrapped in instana.TracingNamedHandlerFunc (or derivates) can not find Instanas Span using opentracing.SpanFromContext.

I think it would be better to use the corresponding opentracing functions (https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L11) so that instana Spans can be retrieved (and augmented) by any client using opentracing-go.

### Minimal, Complete, Verifiable, Example

-

### Go Version

go1.18

### go.mod

```shell
-
```

### go env

```shell
-
```

willianpc:
Hey @martinei thanks for bringing this topic up.

Let us take a look at it and see what it is possible to be done and we come back to you.

Best,
_ Willian

martinei:
HI @willianpc 
Here is an pull request with an unittest: https://github.com/instana/go-sensor/pull/344

The "real world" scenario is to keep most of the code instana agnostic.

martinei:
I now see you already have som very similair MR. I am not sure what information exactly our are missing.

willianpc:
Thanks for the explanation, @martinei , and also for the PR. As you pointed out, we already have a draft PR with the same content.
The reason behind the question was to be sure that we understood your motivation and, consequently, provide a proper change.

Once we have a final PR merged, I reply to this thread with the link.

willianpc:
Hello @martinei ,

After a discussion with the team, we concluded that this change - although simple and viable - can cause unexpected issues to some clients. eg: Lost of span correlation.

The way the the tracer is currently designed generates a tight relationship with the Opentracing API, which prevents us from modifying some parts of the code without the risk of having breaking changes, which is unfortunately the case here.

We really appreciate your suggestion and would like to implement it in a future new major release of the tracer, where usually break changes are expected.

Thanks!

_ Willian

PR #342, opened 2022-09-12
replacing Instana specific context key by opentracing

willianpc:


willianpc:
We are not moving forward with this change due to potential breaking changes.
However this change is recommended for a future new major version of the tracer: https://github.com/instana/go-sensor/issues/340#issuecomment-1247696737

PR #347, opened 2022-09-19
Instrumenting sql.Out

willianpc:
## What
This PR enhances the SQL instrumentation by implementing the [driver.NamedValueChecker](https://pkg.go.dev/database/sql/driver@go1.19.1#NamedValueChecker) interface.

## Why
When a stored procedure with an "out" (`sql.Out{...}`) parameter is provided, the sql package will try to [validate special argument types](https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/database/sql/convert.go;l=124).
This validation is optional by the driver developer, or else sql will fallback to a [default checker](https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/database/sql/convert.go;l=166).

## Attention points
In some drivers, there may be the case where a driver specific checker will not be called if only the Statement (`driver.Stmt`) implements the method `CheckNamedValue(d *driver.NamedValue) error`. We need to document this case in our public docs.

Although this PR fixes the problem with DB2 AND instruments stored procedures, we should instrument  [ibmdb](https://github.com/ibmdb/go_ibm_db) in the future.

### DIFF
--- a/instrumentation_sql.go
+++ b/instrumentation_sql.go
@@ -231,20 +231,7 @@ func (conn *wrappedSQLConn) CheckNamedValue(d *driver.NamedValue) error {
 		return c.CheckNamedValue(d)
 	}
 
-	// If driver.Conn doesn't implement the CheckNamedValue method, we attempt to borrow the method from driver.Stmt.
-	stmt, err := conn.Prepare("select name from table where name = ?")
-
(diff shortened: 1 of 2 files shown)


andriikushch (approved):
LGTM! 👍 Good job!

PR #348, opened 2022-09-19
Add region to the dynamo db span

andriikushch:
This PR updates AWS SDK instrumentation. It extends DynamoDB span with a "region" tag.

### DIFF
--- a/instrumentation/instaawssdk/dynamodb_tags.go
+++ b/instrumentation/instaawssdk/dynamodb_tags.go
@@ -11,47 +11,52 @@ import (
 )
 
 func extractDynamoDBTags(req *request.Request) (opentracing.Tags, error) {
+	var tags opentracing.Tags
 	switch params := req.Params.(type) {
 	case *dynamodb.CreateTableInput:
-		return opentracing.Tags{
+		tags = opentracing.Tags{
 			dynamodbOp:    "create",
 			dynamodbTable: aws.StringValue(params.TableName),
-		}, nil
+		}
 	case *dynamodb.ListTablesInput:
-		return opentracing.Tags{
+		tags = opentracing.Tags{
 			dynamodbOp: "list",
-		}, nil
+		}
 	case *dynamodb.GetItemInput:
-		return opentracing.Tags{
+		tags = opentracing.Tags{
 			dynamodbOp:    "get",
 			dynamodbTable: aws.StringValue(params.TableName),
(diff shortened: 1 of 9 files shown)


instrumentation/instaawssdk/dynamodb_tags.go
```diff
@@ -11,47 +11,52 @@ import (
...
+		tags = opentracing.Tags{
 			dynamodbOp:    "scan",
 			dynamodbTable: aws.StringValue(params.TableName),
-		}, nil
+		}
 	default:
 		return nil, errMethodNotInstrumented
 	}
+
+	tags[dynamodbRegion] = aws.StringValue(req.Config.Region)
```
basti1302, line 59: You could extract capturing the `dynamodbTable` tag also from the individual cases and put it here, only omitting it if type is "list tables". Just a thought.

basti1302 (approved):
LGTM

instrumentation/instaawssdk/dynamodb_tags.go
willianpc: +1 for this suggestion, although feel free to leave it to another time if you want, as it does not directly impact the change being done.

willianpc (approved):
LGTM.
My only comment is about Bastian's suggestion.

willianpc:
> LGTM
> 
> (CircleCI is not happy though)

Just for clarity, this error happens because this PR depends on a new version (v1.44.0) of the tracer core.
Right now, this PR points to an unknown version of the tracer core. 
Once Andrii merges that PR and releases the aforementioned v1.44.0, this PR will then find the correct version - which it depends on :)

instrumentation/instaawssdk/dynamodb_tags.go
andriikushch: Good point 👍 

I played with code a little bit, but so far I do not have an idea how to do this without an extra if statement :(. In my opinion, having it doesn't make code easier to read.

basti1302: Yes, that would need an extra if statement. It was really just a thought to reduce the code duplication a bit. But I agree, it is totally a matter of taste, so feel free to ignore the suggestion entirely. 👍

andriikushch: 👍

PR #349, opened 2022-09-28
Update dependencies

andriikushch:
This PR updates instrumentations to use the latest go-sensor version.


willianpc (approved):
LGTM
That was a nice initiative, to update the tracer in each instrumented lib.

PR #354, opened 2022-10-12
chore(sqs, sns): remove fallback to legacy headers

basti1302:
This is the last step of a multi-year header migration from using X_INSTANA_T/X_INSTANA_S with signed long values over using X_INSTANA_ST/X_INSTANA_SS with string values back to X_INSTANA_T/X_INSTANA_S with string values.

In this final step, we remove the fallback to the legacy header names X_INSTANA_ST/X_INSTANA_SS on the consumer side. Tracers have been sending X_INSTANA_T and X_INSTANA_S with string values since approximately January 2020, so the grace period for updating has been a generous 22 months.

### DIFF
--- a/instrumentation/instaawssdk/propagation_test.go
+++ b/instrumentation/instaawssdk/propagation_test.go
@@ -45,26 +45,6 @@ func TestSpanContextFromSQSMessage(t *testing.T) {
 				},
 			},
 		},
-		"legacy keys": {
-			MessageAttributes: map[string]*sqs.MessageAttributeValue{
-				"Custom": {
-					DataType:    aws.String("String"),
-					StringValue: aws.String("custom attribute"),
-				},
-				"X_INSTANA_ST": {
-					DataType:    aws.String("String"),
-					StringValue: aws.String("00000000000000010000000000000002"),
-				},
-				"X_INSTANA_SS": {
-					DataType:    aws.String("String"),
-					StringValue: aws.String("0000000000000003"),
-				},
-				"X_INSTANA_SL": {
-					DataType:    aws.String("String"),
-					StringValue: aws.String("1"),
-				},
-			},
-		},
 	}
 
 	for name, msg := range examples {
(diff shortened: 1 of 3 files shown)


instrumentation/instaawssdk/propagation_test.go
```diff
@@ -86,7 +66,7 @@ func TestSpanContextFromSQSMessage(t *testing.T) {
 	})
 }
 
-func TestSpanContextFromSQSMessage_LegacyHeaders(t *testing.T) {
+func TestSpanContextFromSQSMessage_2(t *testing.T) {
```
willianpc, line 69: Hm, good catch.
As far as I can see, there is nothing here related to the legacy headers, and also, this test seems to be already covered in the original `TestSpanContextFromSQSMessage`.
@sanojsubran would you be so kind and double check, just to be sure?
It sounds to me that we can remove this test completely.

willianpc: But this is not related to this PR, so the removal (or not) of this test should not prevent the PR to be approved.

willianpc (approved):
LGTM.
Thanks for the effort at updating our tracer.
I added a small comment, not related to the change, about a potential unnecessary test. But this should not prevent the PR to move forward.

instrumentation/instaawssdk/propagation_test.go
basti1302: Right, I meant to ask you folks about this test when putting the PR up, but then I forgot about that.

AFAICT it really has nothing to do with legacy headers. I tried to check the git history to see if this test has ever been about legacy headers but AFAICT it was just misnamed from the start.

If you both think that it's covered in the other test (I guess it is) then let's just remove the test in this PR. If, on the other hand, it covers some aspect that is not covered in the other test then we should probably find a better name instead of `_2` :-) 

Let's wait for Sanoj's input and then we can decide.

sanojsubran: I agree with you guys. It seems to me also that we can remove it.

sanojsubran (approved):
LGTM

instrumentation/instaawssdk/propagation_test.go
basti1302: done

PR #355, opened 2022-10-18
fix checkHost function for fsm

andriikushch:
Fix bug introduced https://github.com/instana/go-sensor/commit/601177eb79ab2090a826b122484ded9b4daae45b.

There was missed request in the case when we try to reach the agent via the gateway address.


sanojsubran (approved):
Looks good

willianpc (approved):
LGTM.
Thanks for applying this fix!

PR #356, opened 2022-10-20
Added ShutdownSensor() function

willianpc:
This PR introduces a new public `instana.ShutdownSensor()` function, responsible for cleaning up the internal global sensor reference.

This is the first phase of a two phase refactoring.

### Why
Although we want the global internal sensor to be available at all times in a real case scenario, the same global sensor compromises our tests, since we are blindly using always the same instance of it, which comes together with an agent client instance, as well as a fsm instance. All with an initial setup.

With a cleanup of this sensor, we can assure that no tests are passing with a false positive, and potentially uncover hidden issues in the tracer.

Later on, once this new function is merged into the master branch and a new release of the tracer is done, the second phase of the refactoring comes in: to update all instrumentations to the new version, and making usage of the new function to cleanup the sensor after each test.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -140,9 +140,6 @@ func (l *Logger) SetPrefix(prefix string) {
 
 // Debug appends a debug message to the log
 func (l *Logger) Debug(v ...interface{}) {
-	l.mu.Lock()
-	defer l.mu.Unlock()
-
 	if l.lvl < DebugLevel {
 		return
 	}
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -65,16 +65,16 @@ type Logger struct {
 // New initializes a new instance of Logger that uses provided printer as a backend to
 // output the log messages. The stdlib log.Logger satisfies logger.Printer interface:
 //
-// 	logger := logger.New(logger.WarnLevel, log.New(os.Stderr, "instana:", log.LstdFlags))
-// 	logger.SetLevel(logger.WarnLevel)
+//	logger := logger.New(logger.WarnLevel, log.New(os.Stderr, "instana:", log.LstdFlags))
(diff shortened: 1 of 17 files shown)


logger/logger.go
```diff
@@ -140,6 +140,9 @@ func (l *Logger) SetPrefix(prefix string) {
 
 // Debug appends a debug message to the log
 func (l *Logger) Debug(v ...interface{}) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
```
sanojsubran, line 145: Query: Do we really need to guard this block? Secondly if it is necessary, we might have to add it for the Info, Warn, Error also.

sensor.go
```diff
@@ -220,6 +220,14 @@ func Flush(ctx context.Context) error {
 	return sensor.Agent().Flush(ctx)
 }
 
+// ShutdownSensor cleans up the internal global sensor reference. The next time that instana.InitSensor is called,
+// directly or indirectly, the internal sensor will be reinitialized.
+func ShutdownSensor() {
+	if sensor != nil {
```
andriikushch, line 226: Would it make sense to protect the `sensor` with mutex?

logger/logger.go
willianpc: This was added due to some race condition issues that happen when we make changes to the instrumentations.
It doesn't make sense to have this change in this PR, but there is a good chance that I will re-add this change in the coming PR, which is, the one that updates every instrumentation to use the latest core version, where the global sensor is removed from the tests.

Then we can discuss the need to add them for all methods. I would initially argue that we should only lock the code where a race condition actually is present, rather than adding them because a similar method needs it. But I would love to discuss together before deciding on one thing or the other :)

andriikushch (approved):
LGTM 👍

PR #360, opened 2022-11-03
Agent Client / State machine separation

willianpc:
This PR decouples the circular relationship between the agent client and the state machine.
It also introduces a couple of medium improvements, like introducing utility functions and a considerable amount of cleaning around some functionalities, as it is the case of the HTTP client communication function - a huge multi purpose function often complex to read and understand, specially while debugging the code.


sanojsubran (commented):
Looks good to me!

andriikushch (approved):
LGTM 👍

PR #362, opened 2022-11-17
Shell script to release all instrumentations

willianpc:
This PR introduces a shell script that helps to update all instrumentations to reference the latest core module being released.
It also provides utilities to release every instrumentation after they are up to date with the latest core module.

#### For instrumentation update

1. Updates every instrumentation's `go.mod` and `go.sum` files by referencing the latest release core module.

#### For instrumentation release

1. Creates a new tag for every instrumentation by incrementing the minor version
2. Updates the `version.go` file of each instrumentation with the new version
3. Commits changes to master branch and push all new tags upstream
4. Releases every single instrumentation

### DIFF
--- a/instrumentations.sh
+++ b/instrumentations.sh
@@ -1,0 +1,98 @@
+#!/bin/sh
+
+# This script provides utilities to update all instrumentations to refer to the latest core module.
+#
+# Usage
+#
+# Update instrumentation phase
+#
+# 1. Run `./instrumentations.sh update`. This will update all instrumentations to refer to the latest core module.
+# 2. Run `make test` to ensure that all instrumentations work with the latest core.
+# 2.1. If any errors are detected in some instrumentation, fix them.
+# 3. Make sure that everything is ok before you commit.
+# 5. If everything is fine, commit your changes, open a PR and get it merged into the master branch.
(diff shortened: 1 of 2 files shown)


andriikushch (approved):
LGTM 👍

instrumentations.sh
```diff
@@ -0,0 +1,93 @@
...
+
+    # Tags to be created after version.go is merged to the master branch with the new version
+    TAGS="$TAGS $LIB_PATH@v$MAJOR_VERSION.$MINOR_VERSION.0"
+  done
+
+  # Commit all version.go files to the master branch
+  git add ./instrumentation/**/version.go
+  git add ./instrumentation/**/**/version.go
+  git commit -m "Bumping new version of the instrumentation"
+  git push origin master
```
sanojsubran, line 71: Do we have any branch protection rules for the master branch which prohibits from direct 'git push' ?

willianpc: We don't. 
Currently, we already create releases from the master branch (where these same commits are used today), although there is nothing in our docs saying that a release should happen from the master branch.

sanojsubran (approved):
LGTM !

PR #366, opened 2022-11-22
Added docker-compose.yaml in the root folder

willianpc:
This PR introduces a `docker-compose.yaml` file in the root of the project, initially exposing only PostreSQL.
The idea is to be able to run integration tests locally, which currently are expected to happen only in Circle CI.

Once this PR is merged, the following steps can be followed to run pgx integration tests locally.

Run these commands from the root folder of the project:

```bash
$ docker-compose up -d postgres
$ make integration
$ docker-compose stop postgres
```

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
 
-  postgre:
+  postgres:
     image: postgres:14
     ports:
       - 5432:5432
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,12 @@
+version: '3'
+
+services:
+
+  postgre:
+    image: postgres:14
+    ports:
+      - 5432:5432
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_DB=postgres
+      - POSTGRES_PASSWORD=mysecretpassword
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+
+services:
+
+  postgre:
```
sanojsubran, line 5: Nitpicking: 'postgre' intentionally ?

sanojsubran (approved):
Looks good

docker-compose.yaml
sanojsubran, line 5:
```suggestion
  postgres:
```

willianpc: Good point. What is the right name?
I checked the official page before writing that: https://www.postgresql.org/
What do you suggest to name it? Maybe although the real name is `postgre` people call it something else?

willianpc: Please, read my reply in the previous comment

sanojsubran: It is absolutely fine if you purposefully wrote it. Kindly ignore my comment.

willianpc: [Wikipedia](https://en.wikipedia.org/wiki/PostgreSQL) tells me that `Postgres` is also correct:

```
PostgreSQL also known as Postgres...
```

I will change it :)

sanojsubran: Got it ! Kindly ignore this.

PR #369, opened 2022-11-22
Better matcher of excluded matches for the instrumentation list

willianpc:
This PR improves the way that instrumentation directories are filtered in the instrumentations.sh script.
There are cases when an instrumentation folder has subdirectories with examples or test code with a go.mod that mistakenly matches the instrumentation list.

This change attempts to offer a more accurate and flexible way to filter unwanted matches.


sanojsubran (approved):
Looks good!

PR #371, opened 2022-11-23
Added replace and dropreplace command options

willianpc:
This PR enhances the `instrumentations.sh` script by adding options that make all instrumentations to use the local core module. They are:

1. **replace**: Adds the `replace` directive into the go.mod file of each instrumentation to reference the local core
2. **dropreplace**: Removes any `replace` directive that changes `github.com/instana/go-sensor` to a different path

No tracer code has been modified in this PR.


sanojsubran (approved):
Looks good!

PR #373, opened 2022-11-29
Race condition improvement

willianpc:
This PR aims to fix race conditions in our unit tests and multiple executions of the same tests where residual instances or variables could be left not cleaned up.
More precisely, this PR assures that tests pass with the arguments `-race -count=2`.

**Integration tests are not part of this task, but they will be tackled soon after.**

**Some code from the core module has been modified. But all changes were adding Lock/Unlock mutexes.**

### Note
Due to the nature of our tracer been split into core module and instrumentations, the `-race -count=2` flags are not enabled in this PR. 

For that, we need to release the core module first, then, in a second step we release all instrumentations referencing the new released core module, where we can finally enable these flags.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -135,8 +135,8 @@ func newAgent(serviceName, host string, port int, logger LeveledLogger) *agentS 
 
 // Ready returns whether the agent has finished the announcement and is ready to send data
 func (agent *agentS) Ready() bool {
-	agent.mu.RLock()
-	defer agent.mu.RUnlock()
+	agent.mu.Lock()
+	defer agent.mu.Unlock()
 
 	return agent.fsm.fsm.Current() == "ready"
 }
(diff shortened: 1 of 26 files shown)


andriikushch (commented):
Hi @willianpc,

In general, fixes look fine. As a suggestion, if would suggest having a closer look at the `sensor.Agent().Ready()` and check if we could keep the state in some `atomic` value, instead of having method invocation every time and necessity to guard everything with extra mutexes.

willianpc:
I will close this PR as it has been hanging here for a while.
We also discussed previously that some race conditions may be false negatives, as we have a global `sensor` which would not cause an issue in a real scenario. Also, we may receive race conditions from the libs we instrument, which would make it nearly impossible for us to solve them.

In short, there is a high probability that we would constantly face race condition issues in the CI tool, which would end up delaying our work and not helping as much as I initially expected.

PR #374, opened 2022-12-02
Agent client for azure

sanojsubran:
This PR adds an agent client for supporting the azure services. This enables pushing the payload to the serverless acceptor and correlates the data to the entityIDs of the packets.


azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+	a.spanQueue = a.spanQueue[:0]
+	a.mu.Unlock()
+
+	for i := range payload.Spans {
+		payload.Spans[i].From = from
+	}
+
+	buf := bytes.NewBuffer(nil)
+	if err := json.NewEncoder(buf).Encode(payload); err != nil {
+		return fmt.Errorf("failed to marshal traces payload: %s", err)
```
willianpc, line 121: In this case the spans are not sent back to the buffer.
Should we add a debug log or something here? So customers will know that some spans will be lost.

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	payloadSize := buf.Len()
+	if payloadSize > maxContentLength {
+		a.logger.Warn(fmt.Sprintf("failed to send the spans. Payload size: %d exceeded max size: %d", payloadSize, maxContentLength))
+		return payloadTooLargeErr
```
willianpc, line 127: Same thing here: should we add a debug to log that spans will be lost?
We don't send them back to the buffer of spans. Which is valid, since the size will remain the same in the next flush attempt, and it would stop here again anyway.

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+	a.spanQueue = append(a.spanQueue, spans...)
+}
+
+func (a *azureAgent) sendRequest(req *http.Request) error {
+	req.Header.Set("X-Instana-Host", a.snapshot.Host)
+	req.Header.Set("X-Instana-Key", a.Key)
+
+	resp, err := a.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send request to the serverless agent: %s", err)
```
willianpc, line 160: Another place where the spans will be lost if the request fails (pls correct me if I am wrong).
If so, maybe we can also add a debug log here to inform about the loss of spans.

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			a.logger.Debug("failed to read serverless agent response: ", err.Error())
```
willianpc, line 168: Not sure what this part here. Would it be relevant to log a debug message about lost spans?

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (a *azureAgent) collectSnapshot() {
+	if a.snapshot.EntityID != "" {
+		return
+	}
+
+	var subscriptionID, resourceGrp, functionApp string
+	if val, ok := os.LookupEnv("WEBSITE_OWNER_NAME"); ok {
+		arr := strings.Split(val, "+")
+		if len(arr) > 1 {
```
willianpc, line 189: Is there any difference if `arr` has more than one element?
My point is, could this be `len(arr) >= 1` instead of `len(arr) > 1` or you have to be sure that the slice has at least 2 elements?

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+
+	if val, ok := os.LookupEnv("WEBSITE_RESOURCE_GROUP"); ok {
+		resourceGrp = val
+	}
+
+	if val, ok := os.LookupEnv("APPSETTING_WEBSITE_SITE_NAME"); ok {
+		functionApp = val
+	}
+
+	entityID := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Web/sites/%s",
```
willianpc, line 202: If any of the tests above at lines 190, 195 and 199 fail, these strings will be empty in the Sprintf.
In this case I assume that `entityID` will be wrong, no? If the entityID is incorrect, the correlation with the infrastructure will not happen. Maybe we could also add a debug log here to inform customers about potential issues in the infra correlation?

azure_agent.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func (a *azureAgent) SendProfiles([]autoprofile.Profile) error { return nil }
+
+func (a *azureAgent) Flush(ctx context.Context) error {
+	a.collectSnapshot()
+
+	if a.snapshot.EntityID == "" {
+		return ErrAgentNotReady
```
willianpc, line 93: Would it make sense to return this test in the `Ready` method?

Something like this:
 ```go
func (a *azureAgent) Ready() bool {
    return a.snapshot.EntityID != ""
}
```

willianpc (approved):
Good job! Overall, it looks good to me.
Most of my comments are about adding extra logs, but you may want to go through all of them as some comments include questions.
I will approve the PR since I don't see anything that prevents us to move forward.
If you feel comfortable with my comments, feel free to go ahead an merge it :)

PR #376, opened 2022-12-12
SQL Instrumentation: Handling drivers that don't implement the CheckNamedValue method

willianpc:
This PR fixes an issue in the SQL instrumentation for when the method `CheckNamedValue` implementation is not found in the driver for both `driver.Conn` and `driver.Stmt` interfaces.
When this scenario is matched, we fallback to the Go standard lib method `driver.DefaultParameterConverter.ConvertValue` to deal with the value check.

### DIFF
--- a/instrumentation_sql.go
+++ b/instrumentation_sql.go
@@ -231,12 +231,12 @@ func (conn *wrappedSQLConn) CheckNamedValue(d *driver.NamedValue) error {
 		return c.CheckNamedValue(d)
 	}
 
-	// stmt, err := conn.Prepare("select name from table where name = ?")
-	stmt, err := conn.Prepare("select * from tbl")
-	// stmt, err := conn.Prepare("select 1")
+	q := "drop table tbl"
+
+	stmt, err := conn.Prepare(q)
 
 	if err != nil {
-		return err
+		sensor.logger.Debug("Database does not support temporary statement: ", q)
 	}
 
 	if s, ok := stmt.(driver.NamedValueChecker); ok {
@@ -293,10 +293,13 @@ func (stmt *wrappedSQLStmt) CheckNamedValue(d *driver.NamedValue) error {
 		return s.CheckNamedValue(d)
 	}
 
-	// return nil
-
 	var err error
 	d.Value, err = driver.DefaultParameterConverter.ConvertValue(d.Value)
+
+	if err != nil {
+		sensor.logger.Debug("Swallowing DefaultParameterConverter error: ", err.Error())
+		return nil
+	}
 
 	return err
 }
--- a/instrumentation_sql.go
+++ b/instrumentation_sql.go
@@ -231,7 +231,21 @@ func (conn *wrappedSQLConn) CheckNamedValue(d *driver.NamedValue) error {
 		return c.CheckNamedValue(d)
 	}
 
-	return nil
+	// stmt, err := conn.Prepare("select name from table where name = ?")
+	stmt, err := conn.Prepare("select * from tbl")
+	// stmt, err := conn.Prepare("select 1")
+
+	if err != nil {
+		return err
+	}
+
+	if s, ok := stmt.(driver.NamedValueChecker); ok {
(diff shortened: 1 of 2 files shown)


instrumentation_sql.go
```diff
@@ -227,11 +227,38 @@ func (conn *wrappedSQLConn) ExecContext(ctx context.Context, query string, args
 }
 
 func (conn *wrappedSQLConn) CheckNamedValue(d *driver.NamedValue) error {
+	defer func() {
```
willianpc, line 230: If any unexpected error occurs, we return error = nil, so we don't break customers code.

instrumentation_sql.go
```diff
@@ -227,11 +227,38 @@ func (conn *wrappedSQLConn) ExecContext(ctx context.Context, query string, args
...
+	}()
+
 	if c, ok := conn.Conn.(driver.NamedValueChecker); ok {
 		return c.CheckNamedValue(d)
 	}
 
-	return nil
+	q := "create table instanatemp (a varchar(1))"
+
+	stmt, err := conn.Prepare(q)
```
willianpc, line 242: If driver.Conn doesn't implement `CheckNamedValue` we attempt to create a statement `driver.Stmt` and check if it implements the method `CheckNamedValue`.

instrumentation_sql.go
```diff
@@ -227,11 +227,38 @@ func (conn *wrappedSQLConn) ExecContext(ctx context.Context, query string, args
...
+		sensor.logger.Debug("Database does not support temporary statement: ", q)
+	}
+
+	defer stmt.Close()
+
+	if s, ok := stmt.(driver.NamedValueChecker); ok {
+		return s.CheckNamedValue(d)
+	}
+
+	d.Value, err = driver.DefaultParameterConverter.ConvertValue(d.Value)
```
willianpc, line 254: If both `driver.Conn` and `driver.Stmt` don't implement `CheckNamedValue` we fallback to Go's standard lib default validation and conversion method.

instrumentation_sql.go
```diff
@@ -227,11 +227,38 @@ func (conn *wrappedSQLConn) ExecContext(ctx context.Context, query string, args
...
+
+	if err != nil {
```
willianpc, line 256: In the case that the Go default method was used and it failed, we log the error but return nil, to prevent customers code to break.
Most likely it should not come here, but if it happens, chances are that we need to cover a new case based on a specific driver. eg: a driver that does conversions of more complex types, other than the ones checked by the standard lib.

instrumentation_sql_test.go
```diff
@@ -349,3 +405,50 @@ type sqlRows2 struct{}
 func (sqlRows2) Columns() []string              { return []string{"col1", "col2"} }
 func (sqlRows2) Close() error                   { return nil }
 func (sqlRows2) Next(dest []driver.Value) error { return io.EOF }
+
+// Driver use case: driver does not implement NamedValueChecker,arg type checking is internal.
+// The idea is to mock pq: https://github.com/lib/pq/blob/8446d16b8935fdf2b5c0fe333538ac395e3e1e4b/encode.go#L31
+
+type sqlDriver3 struct{ Error error }
```
willianpc, line 412: Yet another mocked driver for a specific case, making the test file more verbose.
Maybe we can try to improve this part. 
Unfortunately, reusing a previous mocked driver with the embedding feature is not so simple, as there are more than one struct involved (driver, conn, statement, result, rows...), and usually each struct will have its own particularity due to the test being made.

instrumentation_sql.go
willianpc: The most important part of the fix is to assure that a converter is evaluated in order to make sure that a value type is valid within the query. eg: providing `int32` to the `pq` driver is invalid and should be converted to `int64`, and so on.

andriikushch (approved):
Good job 👍 

Please return `err` from the default checker. And let's give mocks proper specific names.

PR #377, opened 2022-12-13
Sql fix experiment 2

andriikushch:


instrumentation_sql_conn.go
```diff
@@ -0,0 +1,30 @@
...
+	sensor      *Sensor
+}
+
+func (conn *wConn) Prepare(query string) (driver.Stmt, error) {
+	stmt, err := conn.Conn.Prepare(query)
+	if err != nil {
+		return stmt, err
+	}
+
+	if stmtAlreadyWrapped(stmt) {
```
willianpc, line 23: Maybe here should be `connAlreadyWrapped`?

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+
+var arrayStmt = []string{
+	"driver.Stmt",
+	"driver.StmtExecContext",
+	"driver.StmtQueryContext",
+	"driver.NamedValueChecker",
+	"driver.ColumnConverter",
+}
+
+func inArray(el string, arr []string) bool {
```
willianpc, line 37: Idea: this function could return the index or -1 if element not found.
Then it serves two purposes in one: if the element exists and it gives you its exact position in the array.
This could be useful, for instance, in the `removeOnceFromArr`. Not sure if it's super useful, so I leave it to you to judge :)

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+	fmt.Println("package instana")
+	fmt.Println("import \"database/sql/driver\"")
+
+	arrayConnCopy := make([]string, len(arrayConn))
+	arrayStmtCopy := make([]string, len(arrayStmt))
+
+	copy(arrayStmtCopy, arrayStmt)
+	copy(arrayConnCopy, arrayConn)
+
+	gen("driver.Conn", "w_conn_", arrayConn)
```
willianpc, line 70: Is there a reason why we have a constant `driverStmt` but not one for `"driver.Comm"`? 
Maybe I will find out as I go through the rest of the code.

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+	gen(driverStmt, "w_stmt_", arrayStmt)
+
+	printFunctionMaps(arrayConnCopy, arrayStmtCopy)
+	printUtil()
+}
+
+// todo: use copy instead of modify input array
```
willianpc, line 77: Are you still planning on applying this todo?

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+func gen(basicTypeForWrapper, prefix string, listOfTheOriginalTypes []string) {
+	// generate all subsets of the interfaces
+	conn := combinations.All(listOfTheOriginalTypes)
```
willianpc, line 80: Maybe we can rename `conn` to `comb` or something like this, no? `conn` seems to be misleading here.

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+	var filteredSubsets [][]string
+
+	for _, v := range conn {
```
willianpc, line 83: Suggestion:

```go
	for _, v := range conn {
		if inArray(basicTypeForWrapper, v) && len(v) > 1 {
			filteredSubsets = append(filteredSubsets, v[1:])
		}
	}
```

internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,371 @@
...
+			v = v[1:]
+			if len(v) == 0 {
+				continue
+			}
+			filteredSubsets = append(filteredSubsets, v)
+		}
+
+	}
+
+	// not really important
```
willianpc, line 95: I think we can remove this comment and keep the sorting, no?
This means that the generated code will look more organized and easier to read, as far as I understand.

internal/bin/sql/sqlgen.go
andriikushch: Decided to keep it as it is

internal/bin/sql/sqlgen.go
andriikushch: Changed

internal/bin/sql/sqlgen.go
andriikushch: It is an idea for improvement.

internal/bin/sql/sqlgen.go
andriikushch: Changed

internal/bin/sql/sqlgen.go
andriikushch: Let's remove the comment

willianpc:
A version forked from this branch was merged to master with the addition of outputting the generated file using a Go template.
The original logic was kept intact. Only a couple of functions created to generate the contents of the file were removed or repurposed.

PR #380, opened 2022-12-27
SQL Instrumentation refactoring

willianpc:
This PR is a fork from https://github.com/instana/go-sensor/pull/377 that makes all output data based on a template.
The goal is to refactor our SQL instrumentation in a way that all possible cases are covered (hence the idea of a generated file).

### Motivation for this change
The way Go SQL drivers work is offering a set of interfaces for Connection and Statement that can be combined in different ways. Driver implementations are free to match a variety of combinations in order to create their driver.

This combination variety also means that internally, Go takes different routes to deal with these combinations.
Up until now, our instrumentation had a fixed number of interfaces instrumented, which has worked fine for the majority of the time. But also it means that we could end up altering the original behavior of a driver (Go internals wise).


internal/bin/sql/sqlgen.go
```diff
@@ -0,0 +1,238 @@
+// (c) Copyright IBM Corp. 2022
```
sanojsubran, line 1:
```suggestion
// (c) Copyright IBM Corp. 2023
```

internal/bin/sql/sql_wrappers.tpl
```diff
@@ -0,0 +1,186 @@
+// (c) Copyright IBM Corp. 2022
```
sanojsubran, line 1:
```suggestion
// (c) Copyright IBM Corp. 2023
```

instrumentation_sql_stmt.go
```diff
@@ -0,0 +1,45 @@
+// (c) Copyright IBM Corp. 2021
+// (c) Copyright Instana Inc. 2020
```
sanojsubran, line 2:
```suggestion
// (c) Copyright IBM Corp. 2023
```

instrumentation_sql_conn.go
```diff
@@ -0,0 +1,30 @@
+// (c) Copyright IBM Corp. 2021
+// (c) Copyright Instana Inc. 2020
```
sanojsubran, line 2:
```suggestion
// (c) Copyright IBM Corp. 2023
```

sanojsubran (approved):
Great work @willianpc ! I appreciate your effort in coming up with this.

PR #382, opened 2023-01-01
instaamqp: move from `streadway/amqp` to `rabbitmq/amqp091-go`

Juneezee:
The [`github.com/streadway/amqp`](https://github.com/streadway/amqp) library is no longer actively maintained. The new libary is now maintained by the RabbitMQ core team under a different package name [`github.com/rabbitmq/amqp091-go`](https://github.com/rabbitmq/amqp091-go).

Reference: https://github.com/rabbitmq/amqp091-go

### Reviewers note

This will require dropping support for Go 1.13, 1.14, and 1.15.


willianpc:
Hi @Juneezee and thanks for the heads up and the PR.

I am afraid that we won't be able to replace a name by the other due to compatibility issues with users who still use the archived lib, as this may cause users code to break (specially because there is a deprecation of older versions of Go in the new lib).

However, we do need to do the switch as you propose, but keep both versions co-existing for a while.

I will create a task for our team to add the new lib along side the streadway one.

Best,
_ Willian

willianpc:
The PR https://github.com/instana/go-sensor/pull/383 adds a new instrumentation based on amqp091 inspired by this PR.
Thanks again, @Juneezee

PR #385, opened 2023-01-05
Support for azure functions

sanojsubran:
This PR adds the instrumentation package for the azure functions in Go


willianpc (approved):
Awesome job, Sanoj!
As always, most of my comments are questions or suggestions.
There are however a few requests you should consider to take a look.

I will let the PR approved, so you can merge it once all changes are made.

PR #389, opened 2023-01-25
Added workaround to remove http.status when span is graphql

willianpc:
This PR removes the HTTP span tag `http.status` when the span operation is renamed to `graphql.server`.

### Motivation
GraphQL queries received by a webserver are HTTP incoming requests that must be "merged" into one single entry span, as we cannot have two entry spans (http and graphql).
Our UI will render a span as HTTP if the `http.status` tag is present, so in the case of a GraphQL span, we must ensure that this status is not provided.
Due to a design limitation, the graphql instrumentation doesn't have access to the `instana.spanS` struct. Plus, we are unable to control the span finishing from the graphql instrumentation, which leads us to add this workaround here.

### DIFF
--- a/instrumentation_http.go
+++ b/instrumentation_http.go
@@ -127,7 +127,18 @@ func processResponseStatus(response wrappedResponseWriter, span ot.Span) {
 			span.LogFields(otlog.Object("error", statusText))
 		}
 
-		span.SetTag("http.status", response.Status())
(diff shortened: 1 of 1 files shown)


sanojsubran (approved):
Hi Willian,
I think we can keep this logic for the time being. I do not have a better solution for this now. Please go ahead with the merge.

PR #391, opened 2023-01-27
[PoC] Splitting concerns between core module and instrumentations

willianpc:
This PoC aims to redesign how instrumentations relate to the core module in such a way that instrumentations can be released faster and reducing coupling between instrumentations and the core module.

**Keep in mind that this is a PoC. Some names are not ideal and some code is duplicated in order to assure that the tracer works both in the old way and the new one. This known duplication is important to assure that we can gradually update the tracer with smaller releases if we ever decide to go for this approach in Production.**

Some highlights of the PoC (so far):

1. The conversion of `instana.spanS` to a public version `instana.InstanaSpan`
1. The conversion of `instana.typedSpanData` to a public version `instana.TypedSpanData`
1. The separation of the HTTP Server instrumentation from the core module to the API instrumentation
1. The introduction of a new concept: dynamically registered spans in the core module. This is supposed to replace the manual hard coded inclusion of new spans into the core module
1. The ability to work with both legacy registered spans and new dynamic registered spans
2. The conversion of the "readXXXTag" functions to be public (currently, only `ReadStringTag`).

### DIFF
--- a/delayed_spans.go
+++ b/delayed_spans.go
@@ -9,15 +9,15 @@ import (
 const maxDelayedSpans = 500
 
 var delayed = &delayedSpans{
-	spans: make(chan *spanS, maxDelayedSpans),
+	spans: make(chan *InstanaSpan, maxDelayedSpans),
 }
 
 type delayedSpans struct {
-	spans chan *spanS
+	spans chan *InstanaSpan
 }
 
 // append add a span to the buffer if buffer is not full yet
-func (ds *delayedSpans) append(span *spanS) bool {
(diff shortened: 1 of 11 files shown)


willianpc:
This was an experiment that we did not have proper time to discuss around it.
We also had some other ideas for the tracer in terms of refactoring which were not yet concluded, but the direction we may take would not match with this experiment.
In short, we may replace the mono repo by one single repo with go packages instead of separate modules.
Even if this idea does not move forward, different types of refactoring we may apply would either clash with the ideas on this PoC or changes would be so big that would invalidate the PoC.

PR #392, opened 2023-02-01
Added instrumentation for the builtin handler

willianpc:
This PR adds support for the [builtin handler from the graphql API](https://github.com/graphql-go/handler).
The handler is an HTTP Handler that wraps the graphql API and provides features such as a GraphQL Playground.


sanojsubran (approved):
Looks good to me ! :)

PR #399, opened 2023-02-13
Adding support for redis v9

sanojsubran:
This PR adds support for the latest version, v9 of [go-redis](https://github.com/redis/go-redis) library.

In order to add support for the v9 version of go-redis, the interface functions defined by the new `redis.Hooks` were updated in the `commandCaptureHook` structure with necessary instrumentation code changes.


instrumentation/instaredis/instaredis_test.go
```diff
@@ -43,31 +45,77 @@ func (p *mockPipeline) Do(ctx context.Context, args ...interface{}) *redis.Cmd {
...
+		dial:       func(ctx context.Context, network, addr string) (net.Conn, error) { return nil, nil },
+		process:    func(ctx context.Context, cmd redis.Cmder) error { return nil },
+		pipeline:   func(ctx context.Context, cmds []redis.Cmder) error { return nil },
+		txPipeline: func(ctx context.Context, cmds []redis.Cmder) error { return nil },
+	}
+}
+
+func (c *mockClient) chain() {
+	for i := len(c.hooks) - 1; i >= 0; i-- {
+		if wrapped := c.hooks[i].DialHook(c.current.dial); wrapped != nil {
```
willianpc, line 86: Are these ifs really necessary?
From what I see, `wrapped` will never be `nil`, but maybe I am missing something.
I also ran a quick test locally with this branch without the conditions and tests still pass normally.

If they are needed, could you kindly explain what is the logic behind it?

instrumentation/instaredis/instaredis.go
```diff
@@ -126,27 +127,32 @@ func (h commandCaptureHook) handleHook(ctx context.Context, cmd redis.Cmder, cmd
...
-	h.handleHook(ctx, cmd, []redis.Cmder{})
-	return nil
+type InstanaRedisClient interface {
+	AddHook(hook redis.Hook)
+	Options() *redis.Options
 }
 
-func (h commandCaptureHook) BeforeProcessPipeline(ctx context.Context, cmds []redis.Cmder) (context.Context, error) {
-	return ctx, nil
+func (h commandCaptureHook) DialHook(next redis.DialHook) redis.DialHook {
```
willianpc, line 135: Could you be so kind and add comments for all exposed methods? Including any existing public ones? I forgot to add them when the original instrumentation was made

willianpc (changes requested):
Good job! Overall, the instrumentation looks good and clean.
I have a few comments and questions that you may want to consider before we move forward.
There is also a few other files we need to add in order to release the instrumentation:
- version.go
- README.md
- Makefile
- LICENSE.md

Cheers!
Most of these files can be simply copy/pasted, or slightly changed.
The only file you may need some extra attention is README.md, but you probably can also reuse the original README.md from redis v8 and replace the old methods by the new ones.

willianpc:
@sanojsubran another question for a very near future: How do we want to release these branches?
Because of the mono repo, so far we relied on Makefiles from both the root of the project and the ones from each instrumentation.
We cannot rely on the Makefiles from the branch, as the updated one should always reside on the master branch.
One idea would be to always pull changes from the master branch to instaredis/v9, and then make a release. But we should have a proper discussion to agree on something.

instrumentation/instaredis/go.mod
```diff
@@ -1,10 +1,19 @@
 module github.com/instana/go-sensor/instrumentation/instaredis
 
-go 1.13
+go 1.18
 
 require (
-	github.com/go-redis/redis/v8 v8.11.4
 	github.com/instana/go-sensor v1.49.0
```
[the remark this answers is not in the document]
sanojsubran: I have updated it with the latest version.

instrumentation/instaredis/instaredis.go
sanojsubran: I have added comments for the methods. But here the `commandCaptureHook` is an unexported type and we are not actually exposing the methods to the user.

instrumentation/instaredis/instaredis_test.go
sanojsubran: Since we are mocking the client, I thought this would give a better understanding of the change in the underlying implementation of `go-redis` and this is taken from the original implementation. From a mock redis client perspective, it doesn't have to chain the middleware if it is returning `nil` for any of the implementation interface methods.

sanojsubran:
@willianpc I will raise another PR for the release related changes. Currently there are some challenges which needs to be resolved.

PR #410, opened 2023-03-06
Graphql subscription: less flaky unit tests

willianpc:
This PR attempts to fix unit tests that were eventually failing.
Essentially, it consists of providing a unique mocked pubsub instance to each unit test, instead of a single instance shared among all tests.
A small refinement was also made in terms of retry waiting time and number of attempts.


sanojsubran (approved):
Looks good !

PR #411, opened 2023-03-10
Fixed RoundTripper example

willianpc:
This PR fixes the RoundTripper example.
It adds a proper call to finish the manually created entry span in order to assure that the trace is generated properly.


sanojsubran (approved):
LGTM !

PR #413, opened 2023-03-28
Added lots of new debug logs

willianpc:
This PR adds more debug logs around the state machine that connects to the agent.


agent_communicator.go
```diff
@@ -173,21 +192,35 @@ func (a *agentCommunicator) sendDataToAgent(suffix string, data interface{}) err
 
 	resp, err := a.client.Do(req)
 
+	if resp == nil {
+		a.l.Debug("Sending data to agent: response nil for URL ", url)
+	}
+
 	if resp != nil {
+		respCode := resp.StatusCode
+		if respCode < 200 || respCode >= 300 {
```
sanojsubran, line 201: Query: Is it better to suggest that there could be an issue based on HTTP status code range ? Also Warn instead of Debug

sanojsubran (approved):
Looks good !

agent_communicator.go
```diff
@@ -45,12 +48,14 @@ func (a *agentCommunicator) serverHeader() string {
 	req, err := http.NewRequest(http.MethodGet, url, nil)
 
 	if err != nil {
+		a.l.Debug("Error creating request while attempting to retrieve the 'Server' header: ", err.Error())
```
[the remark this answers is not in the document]
willianpc: > Query: Don't we have to report it as error instead of debug?

No. This function only reports a string, expected to be the value of the header key `Server` or empty string. 
To be honest I don't expect this error to ever happen, but since http.NewRequest reports an error I took the chance to log it just in case.

agent_communicator.go
willianpc: as per our talk, we don't still have a proper handling here, as there is no specific spec to handle http codes returned from the agent here, and this log is purely for troubleshoot purposes (for now atleast)

PR #414, opened 2023-04-04
Introducing the instana.Collector

willianpc:
This PR introduces `Collector`, a type that is composed by the same methods of Sensor, ot.Tracer and LeveledLogger, plus a few convenient methods. Collector is an implementation of the newly introduced `TracerLogger` interface.
It also deprecates `InitSensor` in favor of `StartMetrics`.

### Why
The name "sensor" is misleading. What our tracer provides is a tracer **collector**.
Our tracer communicates to the Go Sensor, which is a plugin of the Instana Agent.

As we cannot delete public APIs, we then deprecate Sensor in favor of using the collector. This will happen in future PRs.

> The sensor instance is still accessible via Collector.LegacySensor() and can be used for instrumentations. Exemple:

```go
c := instana.InitCollector(&instana.Options{
	Service: "my-go-app",
})

handler := func(w http.ResponseWriter, r *http.Request) {
	fmt.Fprint(w, "Ok")
}

http.HandleFunc("/foo", instana.TracingHandlerFunc(c.LegacySensor(), "/foo", handler))
```
### Features

 * The Collector is a concrete implementation of a new interface `TracerLogger`.
 * `SensorLogger` is a small interface representing all non-deprecated methods from `Sensor`
 * The `TracerLogger` interface embeds `Tracer` and `LeveledLogger` and the newly introduced `SensorLogger`.
 * It works as both Tracer and Logger, so their usage is less verbose:
    * Before: sensor.Tracer().StartSpan(...)
    * After: c.StartSpan...()
    * Before: s.Logger().Error()
    * After: c.Error()
 * Usage is simpler. There is no need to provide a tracer and then provide the options. The function InitCollector expects the options as an argument
 * A new `Recorder` option is available. It's usually used for testing, so customers shouldn't worry about setting it.
 * A new method `instana.StartMetrics` should be used in favor of `instana.InitSensor` for better naming.

### DIFF
--- a/adapters.go
+++ b/adapters.go
@@ -14,6 +14,14 @@ import (
 	otlog "github.com/opentracing/opentracing-go/log"
 )
 
+var _ TracerLogger = (*Sensor)(nil)
+
+type SensorLogger interface {
+	Tracer() ot.Tracer
+	Logger() LeveledLogger
+	SetLogger(l LeveledLogger)
+}
+
 // SpanSensitiveFunc is a function executed within a span context
 //
 // Deprecated: use instana.ContextWithSpan() and instana.SpanFromContext() to inject and retrieve spans
--- a/adapters.go
+++ b/adapters.go
@@ -32,6 +32,8 @@ type Tracer interface {
 	Options() TracerOptions
 	// Flush sends all finished spans to the agent
 	Flush(context.Context) error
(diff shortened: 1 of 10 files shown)


sanojsubran (approved):
Looks good to me ! Great work @willianpc !

PR #425, opened 2023-05-22
Using TracerLogger instead of *Sensor in the core module

willianpc:
This PR updates the core module to replace references to instana.Sensor by the instana.TracerLogger interface.
The following code was affected:

* For HTTP instrumentation
* For SQL instrumentation
* SQL generator template updated


sanojsubran (approved):
LGTM !

PR #426, opened 2023-05-24
Using TracerLogger instead of *Sensor in all instrumentations

willianpc:
This PR replaces the reference for `instana.Sensor` in all instrumentations by the `instana.TracerLogger` interface.


instrumentation/instaamqp091/go.mod
```diff
@@ -1,17 +1,10 @@
 module github.com/instana/go-sensor/instrumentation/instaamqp091
 
-go 1.18
+go 1.16
 
 require (
-	github.com/instana/go-sensor v1.53.0
+	github.com/instana/go-sensor v1.55.0
 	github.com/opentracing/opentracing-go v1.1.0
```
sanojsubran, line 7:
```suggestion
    github.com/opentracing/opentracing-go v1.2.0
```

Query: Don't we have to update the opentracing package here ?

instrumentation/instaamqp091/go.mod
```diff
@@ -1,17 +1,10 @@
 module github.com/instana/go-sensor/instrumentation/instaamqp091
 
-go 1.18
+go 1.16
```
sanojsubran, line 3: Query: Why did we move back to go 1.16?

sanojsubran (approved):
LGTM ! Thanks for taking up this tedious task of updating all the libraries !

instrumentation/instaamqp091/go.mod
willianpc: Good catch.
This is a leftover for when I was trying to apply Go workspaces. I will undo the change.

PR #428, opened 2023-05-31
Update instrumentations.sh script to handle v2 subfolders properly

willianpc:
This PR updates the instrumentations.sh utility to handle instrumentations with multiple major versions.
A while ago we introduced the instrumentations of new versions of packages. With that, we added a subfolder under the instrumentation folder with the new major version v2, v3 and so on.
When running the script two problems were happening:
 1. The v1.x.x major version was being overwritten by the v2 one
 2. A tag with the wrong name was being created. eg: instrumentation/instaredis/v2/v2.x.x

With these changes, we make sure that the latest tags for each major version is properly detected, and not only the very latest version is returned (before we got v2.x.x, now we get the proper v1.x.x and v2.x.x).
Tags are also formatted properly without adding `/v2/` in the name, which causes the Go website not to be able to recognize the package.


sanojsubran (approved):
Looks good ! Great job !

PR #432, opened 2023-06-23
Artifactory poc

willianpc:


willianpc:
We are dropping this PoC as a potential implementation for the tracer because we are not supposed to read the Body.
We never know what kind of content is posted in the Body nor even its size, which could cause harm to the application.

PR #434, opened 2023-06-29
Added filter for ignoring the script based release notifications

sanojsubran:
This PR adds a filter for muting the notifications when one updates the instrumentations to the latest core module via the script. This will help in reducing the noise in the channel notifications which can cause otherwise.


willianpc (approved):
LGTM. thanks for taking care of this improvement!

PR #435, opened 2023-07-03
Removed automatic service name set when env var is not provided

willianpc:
This PR removes a recently implemented feature that automatically set the service name based on the executable file when the environment variable INSTANA_SERVICE_NAME was not provided.
According to our specification, the service name should not be set automatically, but explicitly set by the user:

```
Tracers CAN set span.data.service to override the default service name extraction rules and set the service name extracted from that span explicitly.
This SHOULD only be done when explicitly configured or triggered by the customer.
```


sanojsubran (approved):
Looks good!

PR #436, opened 2023-07-05
Added new test for service name not provided

willianpc:
This PR adds a test the confirms that no service name is provided to the span JSON payload when no service is provided via options or environment variable.


sanojsubran (approved):
Looks good

PR #444, opened 2023-07-31
Added support for the instrumentation of fiber package

sanojsubran:
This PR adds support for the instrumentation of fiber package using instafiber package.


willianpc (approved):
Looks pretty good to me!
This was a great job, Sanoj!

PR #445, opened 2023-08-02
Updated the Slack notification action

sanojsubran:
This PR simplifies the slack notification action. Redundant checks for specific release types have been removed since it is taken care in the new build workflow scripts.


willianpc (approved):
LGTM

PR #446, opened 2023-08-15
SQL Driver instrumentation: Database auto detection

willianpc:
This PR modifies the generic SQL driver instrumentation to provide DB specific data for some databases:
 * Postgres
 * MySQL
 * Redis

In addition to it, it registers MySQL spans as we currently have no specific MySQL driver instrumentation, except for the generic SQL instrumentation one.

A few other improvements are also added to the general generic SQL instrumentation:

 * Errors are now reported in the span when they occur.


sql_execer.go
```diff
@@ -18,10 +18,13 @@ type wExecer struct {
 func (conn *wExecer) Exec(query string, args []driver.Value) (driver.Result, error) {
```
sanojsubran, line 18: Query: I see that startSQLSpan() was moved to post the actual query interfaces. Wouldn't this cause inaccurate database query execution durations?

sanojsubran (commented):
LGTM ! As usual good job @willianpc ! :)

sql_execer.go
willianpc: > Query: I see that startSQLSpan() was moved to post the actual query interfaces. Wouldn't this cause inaccurate database query execution durations?

I am not sure if I understood the question.
Moved from where to where?
What do you mean by actual query interfaces? 
The idea was to make the function compliant to detect specific databases. As far as I can see there was no change on how startSQLSpan() effectively works, but I may have overlooked something.
Could you please elaborate?

sanojsubran: As discussed, the span should start before the operation.

willianpc: Yes, thanks for the clarification. That was a very good catch.
Changes are ready for another round of review :)

sanojsubran (approved):
Looks good ! 👍🏽

Issue #447, opened 2023-08-21
Shopify has transferred the ownership of sarama to IBM


VipulZopSmart:
### Problem Description

Hi team,

We used to use this https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instasarama for the instana integration with our code, but now as the ownership has been changed for the sarama, how much time will we be able to get the updated one?
https://github.com/IBM/sarama
Any idea? Any thoughts?

### Go Version

go1.19

### go.mod

```shell
.
```

### go env

```shell
.
```

sanojsubran:
Thanks @VipulZopSmart for reporting the issue. We are planning to provide a release based on the latest version of sarama. We'll update this ticket once it is done.

sanojsubran:
Hi @VipulZopSmart ,
We have released a new version on instasarama v1.16.0 with the relevant changes. Kindly check it [here](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instasarama@v1.16.0).

PR #448, opened 2023-08-22
Support for IBM Sarama v1.41.3

sanojsubran:
This PR contains the changes for moving from Shopify/sarama to. IBM/sarama package. 

This is done in response to the issue reported in [https://github.com/instana/go-sensor/issues/447](https://github.com/instana/go-sensor/issues/447)


sanojsubran:
Interfaces have been modified which is affecting the test cases. This will need more effort.

sanojsubran:
Added changes for testcases.

willianpc (approved):
LGTM!

PR #449, opened 2023-08-30
Marshal map issue

willianpc:
This PR makes sure that maps received from public methods are properly copied internally to prevent modifications that may interfere with marshalling.

Maps are implicitly pointers so the best approach is to create a new map, iterate over the original one and feed the new one.
In the vast majority of cases, this should not be a problem, but it could be an issue in some cases, for instance:

1. If a map is passed ahead but the original one changes in a go routine. 
1. If a variable is of type `interface{}` and it ends up being a map, it becomes really hard to track it, unless we use the reflect package

In our case, `ot.Tags` is a type on top of `map[string]interface{}` so you can have maps as items of `ot.Tags`. We should keep this scenario also in mind for similar issues in the future.

### DIFF
--- a/example/http-database-greeter/go.mod
+++ b/example/http-database-greeter/go.mod
@@ -7,3 +7,5 @@ require (
 	github.com/lib/pq v1.8.0
 	github.com/opentracing/opentracing-go v1.2.0
 )
+
+replace github.com/instana/go-sensor => ../../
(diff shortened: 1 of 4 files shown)


sanojsubran (approved):
LGTM ! Nice work @willianpc  🥇

PR #453, opened 2023-09-20
proposal to discuss with Sanoj

willianpc:


sanojsubran:
It looks good to me.

willianpc:
The sole purpose of this PR was to open a discussion about some ideas that could potentially be interesting for the AWS SDK 2 instrumentation.
We are past that discussion so the PR can be closed.

PR #465, opened 2023-10-17
Database Autodetection for gorm

willianpc:
This PR makes the database auto detection enabled for gorm instrumentation.


sanojsubran (approved):
LGTM !

PR #466, opened 2023-10-17
SQL instrumentation: public StartSQLSpan added

willianpc:
This PR introduces a public version of the private startSQLSpan function.
The idea is to reuse the database auto detection features of this function to be used by instrumentation packages, like is the case of instagorm.
Also as part of this change, the previously added errKey string for each auto detected database has been repurposed as the dbKey only. This allows us to use this dbKey (eg: `pg`, `sqlite`, `mysql`) as a prefix for other tags. eg: `pg.error` or `mysql.stmt` and so on.

Additionally, a few minor additions have been made:

1. A working example of instagorm with sqlite
2. A working example of instagorm with postgres
3. Adaption of CI pipeline accordingly to support these new examples
4. The postgres database provided in docker-compose.yaml now creates a simple table for testing


sanojsubran (approved):
LGTM ! Nice work 👍🏽

PR #470, opened 2023-10-18
Update Instagrpc to grpc library to v1.59.0

sanojsubran:
This PR updates dependent grpc library of Instagrpc to 1.59.0. This is in response to the [bug](https://github.com/instana/go-sensor/issues/469) raised in Github.


willianpc (approved):
LGTM!
Thanks for adding this change :)

PR #481, opened 2023-11-03
Fixing gorm postgresql example code

nithinputhenveettil:
This PR fixes the `relation "students" does not exist` error in the gorm postgresql example code.
Nothing fancy, just fixing the table name and column names in the model struct.

*Changes*
- Implemented the `tabler` interface from gorm, so that the correct database name will be returned instead of the gorm's default one.
- Used gorm tags to specify the primary key and column names according to the DDL statement in the init sql file.
- Removed the `gorm.Model` from the student model as it contradicts with the DDL statement.


willianpc (approved):
LGTM man.
Thanks a bunch for fixing this!

PR #483, opened 2023-11-07
exp1

willianpc:


willianpc:
Just some general comments that I noticed while reading the code, so I don't forget it :D 

1. Version of Go should be the lowest as possible, so more customers can use the instrumentation. Usually the same version you see in the instrumented package's go.mod file :)
2. When calling `InstrumentAndConnect`, we should have an option to provide a Tracer. It could be the case that the customer already has one, and both tracers should coexist.
3. Speaking of which, a minor suggestion is to change the name to `Connect`. Because customers, who already don't wanna change their code, only need to add "insta" in front of `gocb.Connect`. IMO, this is a good rule of thumb practice we should have. To make the SDK "UX" as easier to use as possible.
4. I am not entirely sure if having the global `var bucketTypeLookup map[string]string` could cause some issue. Would it make sense to have it as part of some of the structs? We can discuss in a call.

If I remember anything else I update this comment. 
Once again, good job!

nithinputhenveettil:
> Just some general comments that I noticed while reading the code, so I don't forget it :D
> 
> 1. Version of Go should be the lowest as possible, so more customers can use the instrumentation. Usually the same version you see in the instrumented package's go.mod file :)
> 2. When calling `InstrumentAndConnect`, we should have an option to provide a Tracer. It could be the case that the customer already has one, and both tracers should coexist.
> 3. Speaking of which, a minor suggestion is to change the name to `Connect`. Because customers, who already don't wanna change their code, only need to add "insta" in front of `gocb.Connect`. IMO, this is a good rule of thumb practice we should have. To make the SDK "UX" as easier to use as possible.
> 4. I am not entirely sure if having the global `var bucketTypeLookup map[string]string` could cause some issue. Would it make sense to have it as part of some of the structs? We can discuss in a call.
> 5. What is the common use case for couchbase? we connect to the db for every usage or we keep a connection alive ? in case we reuse a connection at all times, we cannot pass the same context always, or it will be a parent span with tons of child spans. otherwise, the way it's being used now is already fine :)
> 
> If I remember anything else I update this comment. Once again, good job!

@willianpc : Thankyou for your review and suggestions.

1, 2, 3. I will make necessary changes in the next commit.
4. This is something I need to check. Actually, I just added a simple in-memory cache for bucket types during development. I'll do a refactoring if needed.
5. Good point on passing the context. I'll clarify in the documentation that users only need to pass the parent span if they want a parent-child relationship. For instance, if it's part of a broader context like a REST API request, then the couchbase span becomes a child span of the HTTP request span. Otherwise, using a nil, TODO, or Background context works perfectly. Alternatively, users can simply omit passing any parent span, and it will still function seamlessly.

willianpc:
dropping this as @nithinputhenveettil is on it

PR #484, opened 2023-11-16
Gocb instrumentation

nithinputhenveettil:
**Changes**

- Added couchbase span type in the core module.
- Added instrumentation for gocb package.
- Added intergration tests for instagocb.


instrumentation/instagocb/go.mod
```diff
@@ -0,0 +1,13 @@
...
+
+require (
+	github.com/couchbase/gocb/v2 v2.6.5
+	github.com/couchbase/gocbcore/v10 v10.2.9
+	github.com/instana/go-sensor v1.58.0
+	github.com/opentracing/opentracing-go v1.2.0
+	github.com/stretchr/testify v1.8.2
+)
+
+replace github.com/instana/go-sensor => ../../
```
nithinputhenveettil, line 13: I'm keeping this replace statement here to prevent test case failures in the current pull request. (As we've implemented some changes in the core module for couchbase span support)

I am planning to remove it in a subsequent pull request before releasing instagocb.

instrumentation/instagocb/README.md
```diff
@@ -0,0 +1,110 @@
+# Instana instrumentation of Couchbase SDK v2 (gocb) for Go
```
nithinputhenveettil, line 1: @instana/eng-go  Please review the README and let me know if there's anything else that should be added.

instrumentation/instagocb/README.md
```diff
@@ -0,0 +1,131 @@
...
+6. Binary Collection
+7. Collection Manager
+8. Collection Data Structures
+   1. List
+   2. Map
+   3. Queue
+   4. Set
+
+> [!NOTE]
+> While you can call methods such as QueryIndexManager or SearchIndexManager from the provided instagocb cluster interface, it's important to note that tracing support for these methods is currently not implemented. If you find the instrumentation of any unlisted features necessary, please feel free to raise an issue.
```
willianpc, line 31: I didn't check the code yet, so maybe you have done this already: 
Would it make sense to implement these methods and just proxing to the original ones and add a debug log message saying they are not implemented? 
Or have this debug message somewhere, eg: when instacodb.Connect is called.
In any case, having this note here is already good enough, tbf.

instrumentation/instagocb/README.md
willianpc: Looks pretty good to me, man!
Very detailed and concise. ⭐⭐⭐⭐⭐

instrumentation/instagocb/instagocb.go
```diff
@@ -0,0 +1,194 @@
+// (c) Copyright IBM Corp. 2023
+
```
sanojsubran, line 2:
```suggestion
// Package Instagocb provides instrumentation for the gocb package
```

Add a line with the package information so that it appears in the godoc

instrumentation/instagocb/README.md
sanojsubran: It looks very detailed and well written. Really good job @nithinputhenveettil !

instrumentation/instagocb/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+
+Compatibility
```
sanojsubran, line 35: Do we need this section?

- gocb version : It will affect if the auto update script updates the instrumentation pacakge.
- go version: This will also add a possible place for error if anyone who updates the package, forgets to update the README.

sanojsubran (commented):
Hi @nithinputhenveettil ,
I had a quick look at the PR and I have added some comments. Kindly check it. I need some time to go through the code and will add further comments if I have any.

instrumentation/instagocb/README.md
nithinputhenveettil: @willianpc : No, I haven't done this yet. 

Initially, I considered doing it, but the issue is that we might have to create all the foundational elements first before proceeding.  For example, developing an interface for QueryIndexManager and writing numerous methods for it (there's a possibility that some of these methods are not instrumented) And for these proxy methods, we may need to create unit/intergration tests as well. 

I'm thinking it might be more efficient to just add a note about this in the readme and call it good. Let me know if you disagree. I don't have any strong reasons for avoiding it, so if you still think it's necessary, we can go ahead and do it. 👍🏾

instrumentation/instagocb/README.md
nithinputhenveettil: Good point. I will remove this from the readme.

instrumentation/instagocb/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/instana/go-sensor/instrumentation/instagocb
```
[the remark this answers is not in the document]
nithinputhenveettil: Good point, I think. We can do it 👍🏾

instrumentation/instagocb/go.mod
```diff
@@ -0,0 +1,13 @@
...
+
+go 1.13
```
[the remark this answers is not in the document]
nithinputhenveettil: @sanojsubran: It looks like they released a new version two days ago. 😞 I initially started this instrumentation to support version 2.6.5, and now they've come out with 2.7.0.

The Go version has been updated from 1.13 to 1.19 in this major release.

In my opinion, we can stick with Go version 1.13, as this is a significant new release from Couchbase, and many users will likely continue using the 2.6 version for an extended period.

@instana/eng-go, what are your thoughts on this?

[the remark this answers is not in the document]
sanojsubran: Oh, Then it is absolutely fine. I was just trying to say that we can keep it consistent with the original package. I was also wondering how the tests went through if the min version is go1.13. Usually something will break since they will be using newer features of Go.

instrumentation/instagocb/go.mod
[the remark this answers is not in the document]
sanojsubran: It indeed will look weird. This is more like an open question from my side.

Angith (approved):

instrumentation/instagocb/go.mod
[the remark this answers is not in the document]
willianpc: At the moment I don't have a strong opinion on this, as we have discussed versioning already in the past with the apparent conclusion that each case can be handled in isolation.
I'd say, go ahead the way you think it's best, and we deal with potential issues as they appear.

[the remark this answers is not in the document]
nithinputhenveettil: I also think it would be better to use "v2" in the module name for some added clarity. I'll make that change now since everyone seems to agree with it.

instrumentation/instagocb/version.go
```diff
@@ -0,0 +1,6 @@
+// (c) Copyright IBM Corp. 2023
+
+package instagocb
+
+// Version is the instrumentation module semantic version
+const Version = "0.1.0"
```
willianpc, line 6: you can leave it as 0.0.0
When you trigger the release GH Action, it will increase the version for you. Otherwise you will get 0.2.0 :)

willianpc (approved):
Awesome job, man.
Quite an experience for the first instrumentation, and you nailed it!
I left a comment about the version, but that's really a tiny thing.

Cheers!

instrumentation/instagocb/go.mod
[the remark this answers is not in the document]
nithinputhenveettil: BTW, if we change the module name to "v2," our release should start from v2.0.0, correct? Otherwise, it might be confusing. Gocb changed their version name to v2 from the v2.0 major release onwards. However, since we are releasing our library for the first time, our version will be 0.1, right? This might seem a bit odd—having "v2" in the module name but the package actually at version 0.1.

@willianpc @sanojsubran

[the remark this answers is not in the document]
sanojsubran: Yes, If we are going with v2, we need to keep the version as 1.0.0 if the file and then do a major version upgrade. That's why I mentioned that this will be a slightly weird thing to do.

instrumentation/instagocb/version.go
nithinputhenveettil: Done.

instrumentation/instagocb/go.mod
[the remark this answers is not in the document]
nithinputhenveettil: If that's the case, I believe we can stick with not adding "v2" to the module name. We'll proceed as is, and as @willianpc suggested, we can address any issues that arise along the way.

PR #486, opened 2023-11-17
Fixing coverage for instapgx

willianpc:
This PR moves the test coverage generation of instapgx after the loop over all other instrumentations.
The reason for that is that the loop also attempts to generate the coverage for instapgx again, and since nothing is tested, the file was being overwritten by the empty one.

### DIFF
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -26,6 +26,11 @@ jobs:
 
             LIB_LIST=$(find ./instrumentation -name go.mod -exec dirname {} \;)
 
+            for lib in $LIB_LIST
+              do echo "Generating test coverage for $lib" && cd "$lib" && go mod tidy && go test -v -coverpkg=./... -cover -covermode atomic -coverprofile coverage.out ./... -json > coverage.json && cd -;
+            done
+
+            echo "Generating test coverage for ./instrumentation/instapgx"
             cd ./instrumentation/instapgx
             go mod tidy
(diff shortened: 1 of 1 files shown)


nithinputhenveettil (approved):
:shipit:

willianpc:
> @willianpc : I have a question. What happens if we have both integration tests and unit tests for one package?
> 
> I mean, what if In the future, if we add some unit tests to the code that already has integration tests, is there a possibility that the unit test coverage might replace the integration test coverage? How about using separate files like coverage.out for unit tests and coverage_integration.out for integration tests?"

In the current scenario the latter will kill the previous one. Maybe there is a way to concat files instead of overwriting them.
On the other hand, this case has never happened as far as I know, and putting effort into fixing this may not be a priority right now. Although I think we should keep this in mind somehow.

nithinputhenveettil:
> > @willianpc : I have a question. What happens if we have both integration tests and unit tests for one package?
> > I mean, what if In the future, if we add some unit tests to the code that already has integration tests, is there a possibility that the unit test coverage might replace the integration test coverage? How about using separate files like coverage.out for unit tests and coverage_integration.out for integration tests?"
> 
> In the current scenario the latter will kill the previous one. Maybe there is a way to concat files instead of overwriting them. On the other hand, this case has never happened as far as I know, and putting effort into fixing this may not be a priority right now. Although I think we should keep this in mind somehow.

I agree, currently, we don't have any Go module with both integration tests and unit tests present, so we may not need to worry about it. However, it might be a good idea to add a small comment or note in the sonarcube.yaml about the potential re-writing case for future reference.

PR #490, opened 2023-11-22
Monitoring and release pipeline

sanojsubran:
This PR intends to add a Github action for monitoring the releases of instrumented packages and creating a PR for the release of new instrumentation packages. This will help to reduce the gap between the supported and latest versions of the packages with minimal manual interventions.


.github/workflows/package_monitor.yml
```diff
@@ -0,0 +1,25 @@
+name: Auto instrumentation for version updates
```
willianpc, line 1: Could the expression "auto instrumentation" cause any kind of serious confusion for readers?
I wonder if people would understand this as something else, like a no-code feature or something like this.
Otherwise, please disregard my comment.

willianpc (approved):
Awesome job, Sanoj.
The script looks pretty clear and self explanatory.
LGTM!

.github/workflows/package_monitor.yml
sanojsubran: How about "Currency"? ;) 

Do you have any suggestion?

willianpc: hahaha funny man!
good question. I did not think of any suggestions. Something around "Auto release for version updates?" or "Auto versioning update for instrumentations". 
It's not a big deal also. We can think of a new name if someone asks about auto instrumentation hehe

sanojsubran: Changed it to "Auto update for instrumentation packages' for the time being.

PR #501, opened 2023-11-28
Gocb instrumentation support for transactions

nithinputhenveettil:
- Added instrumentation for transaction specific methods in `gocb`.
- Added integration tests.

### DIFF
--- a/instrumentation/instagocb/README.md
+++ b/instrumentation/instagocb/README.md
@@ -114,10 +114,141 @@ Sample Usage
 ```
 [Full example][fullExample]
 
+
+Transactions
+------------
+- Get a full idea of using transactions in Couchbase using Go SDK from [here][gocb-transactions-example].
+- Create a new transactions instance by calling `cluster.Transactions()`. Like all other instrumented features, this one also returns an instagocb provided interface (`instagocb.Transactions`) instead of the original one (`*gocb.Transactions`).
+- You can use the same `transactions.Run()` method to start transactions.
+```go
+	// Starting transactions
+	transactions := cluster.Transactions()
+```
+- Commit, Rollback, and all other transaction-specific things are handled by gocb only. The advantage is that you will be getting Instana tracing support on top of transactions. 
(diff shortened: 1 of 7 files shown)


instrumentation/instagocb/transaction_attempt_context.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Insert will insert a new document, failing if the document already exists.
+func (itac *instaTransactionAttemptContext) Insert(collection *gocb.Collection, id string, value interface{}) (*gocb.TransactionGetResult, error) {
+	var tracectx gocb.RequestSpanContext
+	if itac.parentSpan != nil {
+		tracectx = itac.parentSpan.Context()
+	}
+
+	span := itac.iTracer.RequestSpan(tracectx, "TRANSACTION_INSERT")
```
nithinputhenveettil, line 91: Instead of just `INSERT` or `GET` I'm putting `TRANSACTION_` as a prefix to distinguish between transaction specific operations to usual operations.

instrumentation/instagocb/README.md
```diff
@@ -114,10 +114,140 @@ Sample Usage
...
+	}
+ ```
+ [Full example][fullExample]
+
+[gocb-transactions-example]: https://docs.couchbase.com/go-sdk/2.6/howtos/distributed-acid-transactions-from-the-sdk.html#creating-a-transaction
+
 [fullExample]: ../../example/couchbase/main.go
 
-<!---
+<!-- -
```
willianpc, line 249: Why do we have spaces here now? Isn't this going to break the automated version check?
Or is the other way around and now you fixed it?

willianpc (approved):
LGTM!

instrumentation/instagocb/README.md
nithinputhenveettil: nah 😄,! 

It seems like an accidental addition occurred in my vscode when I pressed (⌘ + /). It added an extra space between the comment lines. 

Although this won't impact the automation check, as our AWK script specifically looks for a line with the prefix "target-pkg-url" and nothing else, I'll revert this change since it wasn't intended. Thanks for catching this!

nithinputhenveettil: Done.

PR #504, opened 2023-11-30
fixing echo version in the readme comment for automation script

nithinputhenveettil:
This PR updates the correct version for echo in the instaecho README. Also added missing comment in instasarama README.


Angith (approved):

willianpc (approved):
nice catch, thanks for this fix!

PR #505, opened 2023-11-30
Added instaredisv2 version in the readme comment for automation script

nithinputhenveettil:
This PR updates missing comment for automation script in instaredisv2 README.


willianpc (approved):
LGTM.
Pls keep in mind that the script is not yet ready for sub directories.
As Sanoj is not around at the moment I cannot tell if this could break the script in any way, so we should be prepared for a rollback :)

nithinputhenveettil:
> LGTM. Pls keep in mind that the script is not yet ready for sub directories. As Sanoj is not around at the moment I cannot tell if this could break the script in any way, so we should be prepared for a rollback :)

> LGTM. Pls keep in mind that the script is not yet ready for sub directories. As Sanoj is not around at the moment I cannot tell if this could break the script in any way, so we should be prepared for a rollback :)

You are right. According to this line, https://github.com/instana/go-sensor/blob/main/version_updater.sh#L76, the script will only check for packages in instrumentation folder, not in the sub folders.

I will close this PR. :) thanks for this finding man.

PR #515, opened 2023-12-13
Cosmos instrumentation

Angith:
**Changes**

- Added Cosmos span type in the core module.
- Added instrumentation for Azure cosmos library.
- Added integration tests for instacosmos


instrumentation/instacosmos/containerClient.go
```diff
@@ -0,0 +1,118 @@
...
+		partitionKey azcosmos.PartitionKey,
+		item []byte,
+		o *azcosmos.ItemOptions) (azcosmos.ItemResponse, error)
+}
+
+type instaContainerClient struct {
+	database    string
+	containerID string
+	endpoint    string
+	T           tracing.Tracer
```
willianpc, line 28: Is there a reason why `T` is in capital letter, in comparison to the other attributes?
Maybe I find the answer by reading the rest of the code.

instrumentation/instacosmos/dataBaseClient.go
```diff
@@ -0,0 +1,64 @@
...
+
+// CreateContainer creates a container in the Cosmos database.
+// ctx - The context for the request.
+// containerProperties - The properties for the container.
+// o - Options for the create container operation.
+func (idc *instaDatabaseClient) CreateContainer(
+	ctx context.Context,
+	containerProperties azcosmos.ContainerProperties,
+	o *azcosmos.CreateContainerOptions) (azcosmos.ContainerResponse, error) {
+	return idc.DatabaseClient.CreateContainer(ctx, containerProperties, o)
```
willianpc, line 63: Why is this method needed? 
Please, correct me if I missed something, but this method comes from the embedded DatabaseClient, and here you are calling it normally, without making use of any value from it.
Could this maybe just be removed and rely on the original `idc.DatabaseClient.CreateContainer` call?

willianpc (commented):
LGTM so far, man.
Keep up with the good work!
I added a few comments, I hope they can be helpful :)

instrumentation/instacosmos/containerClient.go
Angith: It doesn't need to be public. I have changed it. Good catch!

instrumentation/instacosmos/dataBaseClient.go
Angith: Yes. I have removed it. Users can directly call the original CreatedContainer method.

instrumentation/instacosmos/client.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+	instana "github.com/instana/go-sensor"
+)
+
+const emptyPrimaryKey string = ""
+
+// Client is the interface that wraps the methods of *azcosmos.Client
+type Client interface {
+	Endpoint() string
+	NewContainer(collector instana.TracerLogger, databaseID string, containerID string) (ContainerClient, error)
```
willianpc, line 22: I think we could make the API leaner for users.
Since we have `instaClient`, which embeds `azcosmos.Client`, maybe we don't need the duplicated `NewDatabase()` and `NewContainer()` methods (our version of them that requires an instance of `instana.TracerLogger`).
Instead, we could have the instance of `instana.TracerLogger` as a private property of instaClient, which would be accessible by `NewDatabase()` and `NewContainer()`. This way you wouldn't need to pass `collector` as an argument, and instead would "overwrite" the original methods from azcosmos, eliminating then the duplicated method implementations.

Users would also benefit from it, as they would have to change less existing code to use our API.

I hope this makes sense. Maybe there is some limitation in the Cosmos SDK that prevents us to make this change. But since we have the Client interface and the instaClient implementation, it should be fine to create the instance of `instana.TracerLogger` only when the client is created.

willianpc (commented):
Awesome job! 

I have partially reviewed the code. As I have made some suggestions that would change some main parts of the instrumentation, I would like to wait for your feedback before I continue
Feel free to ping me on Slack if you see fit :)

instrumentation/instacosmos/client.go
willianpc: > I think we could make the API leaner for users. Since we have `instaClient`, which embeds ...

An example of what I am trying to explain is the instrumentation of redigo:

Here you can see sensor is part of the struct:
https://github.com/instana/go-sensor/blob/main/instrumentation/instaredigo/instaredigo.go#L24

Here you see the sensor being sent to a new instance of the struct (among other properties):
https://github.com/instana/go-sensor/blob/main/instrumentation/instaredigo/instaredigo.go#L46

Here you see the sensor from the struct being used. Note that `Send` does not require a reference to the sensor, as both are properties of the struct :)
https://github.com/instana/go-sensor/blob/main/instrumentation/instaredigo/instaredigo.go#L187

Makefile
```diff
@@ -48,9 +52,12 @@ instrumentation/% :
 	printf "VERSION_TAG_PREFIX ?= $@/v\nGO_MODULE_NAME ?= github.com/instana/go-sensor/$@\n\ninclude ../../Makefile.release\n" > $@/Makefile
 	printf '// (c) Copyright IBM Corp. %s\n// (c) Copyright Instana Inc. %s\n\npackage %s\n\nconst Version = "0.0.0"\n' $(shell date +%Y) $(shell date +%Y) $(notdir $@) > $@/version.go
 
-fmtcheck:
-	@gofmt -l .
-	@test -z $(shell gofmt -l . && exit 1)
+fmtcheck: opisfmt $(MODULES)
```
willianpc, line 55: Is this bit still needed? 
It looks to me that we also need a rollback here..
The code from main branch is:

```ssh
fmtcheck:
	@gofmt -l .
	@test -z $(shell gofmt -l . && exit 1)
```

Could you please check?

example/cosmos/main.go
```diff
@@ -0,0 +1,171 @@
...
+func main() {
+	http.HandleFunc("/cosmos-test", instana.TracingHandlerFunc(collector, "/cosmos-test", handler))
+	log.Fatal(http.ListenAndServe("localhost:9990", nil))
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+
+	erStr := r.URL.Query().Get("error")
+	needError := erStr == "true"
+	if needError {
```
willianpc, line 71: I think it's safe to remove this if statement. We are already setting needError to either true or false on line 70 :)

example/cosmos/main.go
```diff
@@ -0,0 +1,175 @@
...
+	http.HandleFunc("/cosmos-test", instana.TracingHandlerFunc(collector, "/cosmos-test", handler))
+	log.Fatal(http.ListenAndServe("localhost:9990", nil))
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+
+	var needError bool
+
+	erStr := r.URL.Query().Get("error")
+	if erStr == "true" {
```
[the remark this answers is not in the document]
Angith: Good Suggestion. Updated it 👍

example/cosmos/main.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		log.Fatal("Failed to create a container client:", err)
+	}
+
+	id := uuid.New().String()
+	spanID := fmt.Sprintf("span-%s", id)
+
+	// Specifies the value of the partition key
+	var partitionKey string
+	partitionKey = fmt.Sprintf("span-%s", spanID)
```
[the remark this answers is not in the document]
Angith: it was a mistake. Fixed it 👍

instrumentation/instacosmos/client.go
Angith: It makes sense to me as well. Good Suggestion. Updated the code with same

instrumentation/instacosmos/insta_cosmos.go
```diff
@@ -0,0 +1,80 @@
+// (c) Copyright IBM Corp. 2024
```
nithinputhenveettil, line 1: More like a suggestion. How about rename this file to `instacosmos.go` for consistency :)

nithinputhenveettil (changes requested):

Makefile
Angith: I have updated the makefile with changes to remove the check for excluded directories. Could you please review it ?

example/cosmos/main.go
Angith: Oh yes. removed it 👍

instrumentation/instacosmos/insta_cosmos.go
Angith: done 👍

example/cosmos/main.go
```diff
@@ -0,0 +1,175 @@
...
+	// setting item options upon creating ie. consistency level
+	itemOptions := azcosmos.ItemOptions{
+		ConsistencyLevel: azcosmos.ConsistencyLevelSession.ToPtr(),
+	}
+
+	itemResponse, err := containerClient.CreateItem(ctx, pk, b, &itemOptions)
+	return itemResponse, err
+}
+
+func getClient(sensor instana.TracerLogger) (instacosmos.Client, error) {
```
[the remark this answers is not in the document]
Angith: Yes. But now I have tweaked the code a little bit and this function uses it.

sanojsubran (commented):
Good job @Angith !

instrumentation/instacosmos/README.md
```diff
@@ -0,0 +1,97 @@
...
+- The aforementioned functions return an interface that offers the functionalities of `azcosmos.Client`. Utilizing this interface, call the `NewContainer` method to acquire the `ContainerClient` interface. This interface encompasses all the operations supported by `azcosmos.ContainerClient`.   
+
+Sample Usage
+------------
+ ```go
+
+    var collector instana.TracerLogger
+	collector := instana.InitCollector(&instana.Options{
+		Service:           "cosmos-example",
+		EnableAutoProfile: true,
```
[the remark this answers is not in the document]
Angith: Yes. removed 👍

instrumentation/instacosmos/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+[Full example][fullExample]
+
+[fullExample]: ../../example/cosmos/main.go
+
+<!---
+Mandatory comment section for CI/CD !!
+target-pkg-url: github.com/Azure/azure-sdk-for-go/tree/main/sdk/data/azcosmos
+current-version: v0.3.6
+--->
```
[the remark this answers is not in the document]
Angith: Done 👍

supported_versions.md
```diff
@@ -25,5 +25,6 @@
 | 23 | Databases | [go-redis v2](https://pkg.go.dev/github.com/redis/go-redis/v9) | [instaredis](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instaredis/v2) | v8.11.4 | v9.4.0 |
 | 24 | HTTP | [beego](https://pkg.go.dev/github.com/beego/beego/v2) | [instabeego](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instabeego) | v2.1.3 | v2.1.5 |
 | 25 | Other | [aws-sdk-go](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2) | [instaawsv2](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instaawsv2) | v1.21.0 | v1.24.1 |
+| 26 | Databases | [azcosmos](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos) | [instacosmos](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instacosmos) | v0.36.0 | v1.0.0 |
```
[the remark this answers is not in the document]
Angith: Good Catch 👍 . Updated the version

instrumentation/instacosmos/client.go
```diff
@@ -0,0 +1,129 @@
+// (c) Copyright IBM Corp. 2023
```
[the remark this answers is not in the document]
Angith: 💯

example/cosmos/main.go
```diff
@@ -0,0 +1,168 @@
...
+	"os"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+	"github.com/google/uuid"
+	instana "github.com/instana/go-sensor"
+	"github.com/instana/go-sensor/instrumentation/instacosmos"
+)
+
+const (
```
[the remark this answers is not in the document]
Angith: Done 👍

instrumentation/instacosmos/partition_key.go
```diff
@@ -0,0 +1,47 @@
...
+	"strconv"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos"
+)
+
+type pk struct {
+	value string
+}
+
+func NewPK(val string) *pk {
```
nithinputhenveettil, line 18: Please add a one-line comment here. I think we can add comments for all exported (capitalized) names. It will be helpful for the documentation.

Also I have a suggestion using `NewPartitionKey` instead of `NewPK` for better clarity.
I think `instacosmos.NewPartitionKey` makes more sense and is more user-friendly.

nithinputhenveettil: @Angith  :
Also, why can't we return the PartitionKey interface instead of *pk? I mean, NewPK is exported, but pk is a private struct. Whoever uses this externally won't be able to do anything with pk, right?

I think it makes more sense to return PartitionKey here. Feel free to correct me if I'm wrong.

nithinputhenveettil (changes requested):

instrumentation/instacosmos/partition_key.go
Angith: NewPK made it public only for internal integration tests. Since Cosmos Library denies access to the PartitionKey, we're also restricting access. But returning the interface is a good idea. Made the changes for it. Please review it.

supported_versions.md
```diff
@@ -25,5 +25,6 @@
 | 23 | Databases | [go-redis v2](https://pkg.go.dev/github.com/redis/go-redis/v9) | [instaredis](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instaredis/v2) | v8.11.4 | v9.4.0 |
 | 24 | HTTP | [beego](https://pkg.go.dev/github.com/beego/beego/v2) | [instabeego](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instabeego) | v2.1.3 | v2.1.6 |
 | 25 | Other | [aws-sdk-go](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2) | [instaawsv2](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instaawsv2) | v1.21.0 | v1.24.1 |
+| 26 | Databases | [azcosmos](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos) | [instacosmos](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instacosmos) | v0.3.6 | v1.0.0 |
```
nithinputhenveettil, line 28:
```suggestion
| 26 | Databases | [azcosmos](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/data/azcosmos) | [instacosmos](https://pkg.go.dev/github.com/instana/go-sensor/instrumentation/instacosmos) | v0.3.6 | v0.3.6 |
```

.github/workflows/coverage_run_and_upload.yml
```diff
@@ -39,15 +42,15 @@ jobs:
...
-            go mod tidy
-            go test -v -tags=integration -coverpkg=./... -cover -covermode atomic -coverprofile $TRACER_PATH/coverage/coverage_$(date +%s%N)_$RANDOM.out ./... -json > $TRACER_PATH/coverage/coverage_$(date +%s%N)_$RANDOM.json && cd -
+            INTEGRATIONS_TESTS=("instagocb" "instapgx" "instacosmos")
 
-            echo "Generating test coverage for ./instrumentation/instapgx"
-            cd ./instrumentation/instapgx
-            go mod tidy
-            go test -v -tags=integration -coverpkg=./... -cover -covermode atomic -coverprofile $TRACER_PATH/coverage/coverage_$(date +%s%N)_$RANDOM.out ./... -json > $TRACER_PATH/coverage/coverage_$(date +%s%N)_$RANDOM.json && cd -
+            for str in ${INTEGRATIONS_TESTS[@]}; do
+              dir=./instrumentation/"$str"
```
nithinputhenveettil, line 48: Feel free to correct if if I'm wrong but I dont think these quotes are needed here. 

```suggestion
              dir=./instrumentation/$str
```

.github/workflows/sonarqube.yml
```diff
@@ -39,4 +39,4 @@ jobs:
 
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
```
nithinputhenveettil, line 42: Can you revert this file to the previous version? I think these change is not needed.

nithinputhenveettil (changes requested):
@Angith : Good job with the Sonarqube action. Added some minor changes.

nithinputhenveettil (approved):
:shipit:

.github/workflows/coverage_run_and_upload.yml
Angith: It's correct. Good catch

supported_versions.md
Angith: done 👍

PR #528, opened 2023-12-21
Fix incorrect routeID when using echo

ATLeanIX:
Comparing only path caused routeID to be the same for both GET and POST handlers. Using both path and method should result in consistent and correct routeID for each handler.


nithinputhenveettil:
@ATLeanIX : Thank you for submitting the fix and the PR. We appreciate your contribution, and our team will review it promptly.

sanojsubran (approved):
LGTM

PR #529, opened 2023-12-26
Updated sonarqube github action

nithinputhenveettil:
**Problem:**
- GitHub does not pass secrets to a runner if the pull request is from a forked repository. Consequently, our SonarQube action fails due to an invalid Sonar secret.

**Proposed solution:**
- Split the SonarQube action into two separate actions:
- The first action will handle the installation of dependencies and the execution of both unit and integration tests. After running the tests, this action will upload the results to an artifact. Notably, this action does not require any secrets, allowing it to succeed even for pull requests from forked repositories.
- The second action will use the workflow_run trigger to initiate the action. This action will download the test cases from the previously generated artifact and update SonarQube accordingly.

Ref : [Github Blog post.](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

> A new trigger workflow_run was introduced to enable scenarios that require building the untrusted code and also need write permissions to update the PR with e.g. code coverage results or other test results. To do this in a secure manner, the untrusted code must be handled via the pull_request trigger so that it is isolated in an unprivileged environment. The workflow processing the PR should then store any results like code coverage or failed/passed tests in artifacts and exit. The following workflow then starts on workflow_run where it is granted write permission to the target repository and access to repository secrets, so that it can download the artifacts and make any necessary modifications to the repository or interact with third party services that require repository secrets (e.g. API tokens).


Angith (approved):

sanojsubran (approved):
LGTM !

PR #546, opened 2024-01-05
Fixing some issues in version updater bash script

nithinputhenveettil:
**Fixes**
- Fixed issue with version update if the lowest and highest supported versions are same.
- Fixed issue with working directory for each instrumentation.


sanojsubran (approved):
Looks good

PR #565, opened 2024-01-09
Updated instagocb version in example

nithinputhenveettil:


example/couchbase/go.mod
```diff
@@ -6,7 +6,5 @@ require github.com/couchbase/gocb/v2 v2.6.5
 
 require (
 	github.com/instana/go-sensor v1.59.0
-	github.com/instana/go-sensor/instrumentation/instagocb v0.0.0-20231107055240-4ac1225b817a
+	github.com/instana/go-sensor/instrumentation/instagocb v0.1.0
 )
-
-replace github.com/instana/go-sensor/instrumentation/instagocb => ../../instrumentation/instagocb
```
sanojsubran, line 12: We can keep the replace directive in example so that it will be tested against the local code

nithinputhenveettil: @sanojsubran : Good point 👍🏾  We can keep the replace directive here. 😊

nithinputhenveettil:
Closing the PR as the change is not needed.

PR #605, opened 2024-02-07
Increasing waiting time in GraphQL unit tests

nithinputhenveettil:
**Changes**
- Increasing waiting time in the GraphQL unit test.

**Why ?**
- We've observed intermittent failures in the GraphQL unit test with the error message "Condition never satisfied!" Without identifying any apparent issues in the code, we are adjusting the waiting time to see if this error persists. If the issue persists, further investigation will be required.

image


sanojsubran (approved):
Ok. Let's see.

PR #645, opened 2024-03-05
Enhance unit test coverage for go sensor

Angith:


integration_test.go
```diff
@@ -60,6 +60,12 @@ func (srv *serverlessAgent) HandleBundle(w http.ResponseWriter, req *http.Reques
 		body = nil
 	}
 
+	if value, ok := os.LookupEnv("NEED_ERROR"); ok && value == "true" {
```
sanojsubran, line 63: Query: Is it possible to remove env wide setting?

lambda_agent.go
```diff
@@ -163,8 +163,9 @@ func (a *lambdaAgent) sendRequest(req *http.Request) error {
 			return nil
 		}
 
-		a.logger.Info("serverless agent has responded with ", resp.Status, ": ", string(respBody))
-		return nil
+		err = fmt.Errorf("serverless agent has responded with %s : %s", resp.Status, string(respBody))
```
sanojsubran, line 166: Suggestion: Better to move to another PR as it has functional implications

sanojsubran (commented):
Please have a look at the comments

instrumentation/instabeego/web_server_handler_test.go
```diff
@@ -64,31 +63,31 @@ func initBeeApp(t *testing.T, r *instana.Recorder) {
 	sleep(serverDepTime)
 }
 
-func TestMain(m *testing.M) {
-
-	instana.InitSensor(&instana.Options{
+func initSensor() {
+	instana.StartMetrics(&instana.Options{
```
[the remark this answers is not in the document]
Angith: yes. Done 👍

integration_test.go
Angith: changed the settings into a tag in spans.

lambda_agent.go
Angith: done 👍

integration_test.go
```diff
@@ -60,6 +61,20 @@ func (srv *serverlessAgent) HandleBundle(w http.ResponseWriter, req *http.Reques
...
+		log.Printf("ERROR: failed to unmarshal serverless agent spans request body: %s", err.Error())
+	} else {
+		if len(root.Spans) > 0 {
+			if root.Spans[0].Data.SDKCustom.Tags.ReturnError == "true" ||
+				root.Spans[0].Data.Lambda.ReturnError == "true" {
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+		}
+	}
```
sanojsubran, line 76: Suggestion: Just combined the if checks.
```suggestion
if err != nil {
		log.Printf("ERROR: failed to unmarshal serverless agent spans request body: %s", err.Error())
	} else {
		if len(root.Spans) > 0 && (root.Spans[0].Data.SDKCustom.Tags.ReturnError == "true" ||
			root.Spans[0].Data.Lambda.ReturnError == "true") {

			w.WriteHeader(http.StatusInternalServerError)
			return
		}
	}
```

instrumentation/instabeego/web_server_handler_test.go
```diff
@@ -64,31 +63,31 @@ func initBeeApp(t *testing.T, r *instana.Recorder) {
 	sleep(serverDepTime)
 }
 
-func TestMain(m *testing.M) {
-
-	instana.InitSensor(&instana.Options{
+func initSensor() {
```
sanojsubran, line 66: Can we change this to `initCollector` to remove the ambiguity ?

sanojsubran (approved):
Looks good. Please check the comments.

instrumentation/instabeego/web_server_handler_test.go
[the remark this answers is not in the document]
sanojsubran: It would be better to rename the parent function also. We do not want to further advertise the term `sensor` in our `tracer` code.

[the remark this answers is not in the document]
Angith: make sense. Updated

instrumentation/instabeego/web_server_handler_test.go
Angith: yes 👍

PR #654, opened 2024-03-08
Added runtime-generated db name for pgx integration tests

nithinputhenveettil:
Added runtime-generated database name for pgx integration tests. This feature would enable multiple pgx tests to run in parallel.


sanojsubran (approved):
LGTM!

PR #664, opened 2024-03-13
added support for opt-in exit spans without entry spans

Angith:
Changes

* Introduce an environment variable INSTANA_ALLOW_EXIT_AS_ROOT. Based on the environment variable, either the exit spans will be sent to the agent or suppressed. 
* Add unit tests for the above changes.


Angith:
Created other PRs for the feature.
https://github.com/instana/go-sensor/pull/684
https://github.com/instana/go-sensor/pull/688

PR #678, opened 2024-03-18
ci: example folder excluded from unit and integration tests

nithinputhenveettil:
This PR adds the "example" folder to the list of excluded directories for CI build commands.


sanojsubran (approved):
Looks good to me

PR #701, opened 2024-04-02
ci: updated the minimum version of go to 1.21 in tekton

nithinputhenveettil:
This PR aims to update the minimum supported Go version to 1.21 for Tekton pipelines.
Ref : #694


sanojsubran (approved):
Looks good to me!

PR #705, opened 2024-04-02
Correction to Go Tracer version details

sanojsubran:
This PR fixes the issue in the support matrix.


nithinputhenveettil (approved):
Nice catch! Thanks for the correction @sanojsubran. ❤️

PR #768, opened 2024-05-13
Modification was made to the Tekton pipeline to exclude tests related to Instacosmos from Go 1.21

nithinputhenveettil:
This PR aims to modify the Tekton pipeline to exclude tests related to` Instacosmos` from Go version 1.21. This is because `azcosmos` has restricted the minimum Go version to 1.22 in their new update.

The [pipeline run succeeded](http://localhost:8001/api/v1/namespaces/tekton-pipelines/services/tekton-dashboard:http/proxy/#/namespaces/default/pipelineruns/update-instrumentations-instacos-go21-88484fc) in Go 1.21 after implementing this change.

### DIFF
--- a/.tekton/ci-build/github-eventlistener.yaml
+++ b/.tekton/ci-build/github-eventlistener.yaml
@@ -60,7 +60,7 @@ spec:
         - name: go-version
           value: "1.21"
         - name: excludeDirs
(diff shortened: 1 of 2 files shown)


sanojsubran (approved):
Looks good !

PR #814, opened 2024-06-18
fix:updated the dependencies for instacosmos in example folder

sanojsubran:


sanojsubran:
@nithinputhenveettil Thanks man for the quick review! :)

PR #825, opened 2024-06-24
doc:added entry for the pgx/v5 in the support matrix table

sanojsubran:
This PR adds an entry in the support matrix table for the pgx v5 instrumentation.


nithinputhenveettil (approved):
:shipit:

PR #850, opened 2024-07-08
fix: DB2 spans are tagged as MySQL spans instead of generic DB spans

Angith:
### Bug
DB2 spans are tagged as MySQL spans instead of generic DB spans.

### Cause

In Go Tracer, Connection strings are used to distinguish between SQL-like databases based on the `database/sql` package. 
As MySQL and IBM Db2 connection strings are very similar, the IBM Db2 driver is identified as MySQL DB.

### Fix

The driver name is also considered to differentiate between MySQL and IBM Db2. 

### Drawback

Currently, only the `go_ibm_db` driver is added as the known driver for IBM Db2. If a different driver or driver name is used, it would still identify the DB as MySQL. We can add other IBM Db2 drivers to the known drivers list on request.

### DIFF
--- a/instrumentation_sql.go
+++ b/instrumentation_sql.go
@@ -297,9 +297,8 @@ func ParseDBConnDetails(connStr, driverName string) DbConnDetails {
 
 	fmt.Println("driver name: ", driverName)
 	if isDB2driver(driverName) {
-		if details, ok := parseDBConnDetailsURI(connStr); ok {
+		if details, ok := parseDB2ConnDetailsKV(connStr); ok {
 			fmt.Printf("ParseDBConnDetails, details: %+v \n", details)
-			details.DatabaseName = "db2"
 			return details
 		}
 	}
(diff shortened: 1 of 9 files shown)


instrumentation_sql.go
```diff
@@ -289,7 +292,14 @@ type DbConnDetails struct {
 	Error        error
 }
 
-func ParseDBConnDetails(connStr string) DbConnDetails {
+func ParseDBConnDetails(connStr, driverName string) DbConnDetails {
+
+	if isDB2driver(driverName) {
+		if details, ok := parseDB2ConnDetailsKV(connStr); ok {
```
sanojsubran, line 298: Please handle the parsing failure case also

instrumentation_sql.go
```diff
@@ -430,6 +453,49 @@ func parseMySQLConnDetailsKV(connStr string) (DbConnDetails, bool) {
...
+		case strings.HasPrefix(fieldNorm, "server="):
+			address := field[len("server="):]
+			for i, v := range strings.Split(address, ":") {
+				switch i {
+				case 0:
+					details.Host = v
+				case 1:
+					details.Port = v
+				}
+			}
```
nithinputhenveettil, line 477:
```suggestion
			addr := strings.Split(address, ":")
			details.Host, details.Port = addr[0], addr[1]
```

nithinputhenveettil (changes requested):
Minor suggestions only.

instrumentation_sql.go
Angith: As port is optional here, the port might not be present in the slice returned by the slice.

instrumentation_sql.go
Angith: Good catch! Done

instrumentation_sql.go
Angith: This code is cleaned up! Please check 😊

instrumentation_sql.go
```diff
@@ -92,11 +95,26 @@ func (drv *wrappedSQLDriver) Open(name string) (driver.Conn, error) {
...
+	if isDB2driver(driverName) {
+		if details, ok := parseDB2ConnDetailsKV(connStr); ok {
+			return details
+		}
+
+		return DbConnDetails{RawString: connStr}
+	}
+
+	return ParseDBConnDetails(connStr)
+
```
sanojsubran, line 115: Nitpick: redundant line break in my opinion
```suggestion
```

instrumentation_sql.go
```diff
@@ -306,6 +324,19 @@ func ParseDBConnDetails(connStr string) DbConnDetails {
 	return DbConnDetails{RawString: connStr}
 }
 
+// isDB2driver checks the driver belongs to IBM Db2 database.
```
sanojsubran, line 327:
```suggestion
// isDB2driver checks whether the driver belongs to IBM Db2 database.
```

instrumentation_sql_test.go
```diff
@@ -231,9 +232,183 @@ func TestOpenSQLDB(t *testing.T) {
...
+			Host:  "localhost:50000",
+			Port:  "",
+			DB:    "sample",
+			User:  "db2inst1",
+			Stmt:  "TEST QUERY",
+			Error: "",
+		}, data.Tags)
+	})
+}
+
```
sanojsubran, line 407: Suggestion: If the above tests have differences only in their data part, you can consider moving them to a table driven tests.

sanojsubran (approved):
Good job @Angith. Thanks for taking this up.

nithinputhenveettil (approved):

instrumentation_sql_test.go
Angith: Good Suggestion! Done 👍

PR #882, opened 2024-08-13
feat: support parallel processing for couchbase integration test

Angith:
**Why**
The couchbase integration test does not support parallel execution as of now which causes random failures in CI/CD.

**What**
Added support for parallel execution of the couchbase integration test.


docker-compose.yaml
```diff
@@ -19,11 +19,28 @@ services:
       - 5671:5671
       - 5672:5672
 
-  couchbase-server-sandbox:
-    image: couchbase/server-sandbox:7.1.1
+  couchbase:
```
nithinputhenveettil, line 22: I think we need to test this new couchbase config in sonarqube github action as well.

nithinputhenveettil (approved):
Looks good to me!

sanojsubran (approved):

docker-compose.yaml
Angith: Yes. I think it will work fine as we start the `couchbase` container using the docker-compose file in GitHub action.

PR #893, opened 2024-09-02
feat: added generic serverless agent

nithinputhenveettil:
This PR aims to add a generic serverless agent code to the tracer SDK. Currently, only traces are supported, and collecting metrics is not yet available.


sanojsubran (approved):
Good job @nithinputhenveettil !

PR #931, opened 2024-10-08
refactor: optimized logic for identifying if the database is Redis from the query

Angith:
The usage of regular expressions imposes performance overhead on SQL span creation. This PR consists of the change for a solution to that. Here a map is used to check if the span is created for the Redis database.


context.go
```diff
@@ -9,9 +9,10 @@ import (
 	ot "github.com/opentracing/opentracing-go"
 )
 
-type contextKey struct{}
+type ContextKey string
```
nithinputhenveettil, line 12: Do we really need to make this type exportable? I think it would be better to keep it internal to the Instana package. For the `AddToContext` and `GetValueFromContext` functions, we could take the key as a `string` and handle the conversion to `contextKey` internally. This approach would make the code cleaner and help prevent any misuse of the type.

context.go
```diff
@@ -28,3 +29,15 @@ func SpanFromContext(ctx context.Context) (ot.Span, bool) {
...
 	return sp, true
 }
+
+func AddToContext(ctx context.Context, key ContextKey, value string) context.Context {
+	return context.WithValue(ctx, key, value)
+}
+
+func GetValueFromContext(ctx context.Context, key ContextKey) string {
+	val, ok := ctx.Value(key).(string)
+	if !ok {
```
nithinputhenveettil, line 39: This code is redundant! The value of `val` will be an empty string even if `ok` is false. So, we can simply return `val` directly.

context.go
nithinputhenveettil: I also think that since we only have two keys supported as `contextKey`, we could rewrite this in a more secure way that restricts the use of only these two keys. Currently, we can add any string type as a key, right?

Additionally, the earlier implementation uses `struct{}` as the underlying type for keys, which helps minimize memory usage since it’s embedded in the context. Given that we only have two keys, we could use `int8` or a similar type instead of string, and leverage `iota` to define each key (such as activeSpanKey and redisKey). This approach not only uses less memory but also allows for easy extension in the future if we decide to add more keys.

nithinputhenveettil (commented):
Good job with the code refactoring, Angith!  🎉 
I have just added a few comments.

context.go
Angith: Good suggestion @nithinputhenveettil 👍 . 

> Do we really need to make this type exportable?

My motive was to use this function in instrumentation packages if required in the future. 

> Given that we only have two keys, we could use int8 or a similar type instead of string, and leverage iota to define each key (such as activeSpanKey and redisKey).

Make sense👍.  I have made the type of `contextKey` to `int8` and used `iota`. The functions are changed to `private` and can be used internally.

context.go
Angith: Good catch 👍. Done!

context.go
```diff
@@ -22,9 +22,5 @@ func ContextWithSpan(ctx context.Context, sp ot.Span) context.Context {
 // span, this method returns false.
 func SpanFromContext(ctx context.Context) (ot.Span, bool) {
 	sp, ok := ctx.Value(activeSpanKey).(ot.Span)
-	if !ok {
```
Angith, line 25: Removed as this code is redundant!

sanojsubran (changes requested):
@Angith Please check my review comments. We can discuss if you have any queries.

instrumentation_sql.go
```diff
@@ -164,20 +246,14 @@ func mySQLSpan(ctx context.Context, conn DbConnDetails, query string, sensor Tra
...
 
-func redisSpan(ctx context.Context, conn DbConnDetails, query string, sensor TracerLogger) ot.Span {
-	qarr := strings.Fields(query)
-	var q string
-
-	for _, w := range qarr {
-		if redisCmds.MatchString(w) {
-			q += w + " "
-		}
+	if cmd == "" {
```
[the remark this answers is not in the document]
Angith: Suppose we identified the database name from the connection URL. In that case, the `cmd` passed would be an empty string. Hence, we need to find out the Redis command.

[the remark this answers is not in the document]
sanojsubran: Ok.I understood.

sanojsubran (approved):

PR #938, opened 2024-10-24
poc: refactoring sql instrumentation

Angith:
This PR optimizes the creation of SQL span tags by decoupling tag generation from query execution. Currently, all span tags—including connection details, which are constants—are generated during query execution. This change moves the creation of these constant tags to the connection setup phase, allowing them to be reused across multiple queries executed through the same connection instance.

Additionally, span creation is now handled by a single object, enabling easier extension and the incorporation of more data in the future. A `Tags` interface has been introduced to streamline the process of adding tags to spans, ensuring efficient tag management. This refactor also aims to improve code structure, enabling cleaner design and facilitating future expansion.

By leveraging a single object for span creation, rather than passing parameters and constructing tags for each query, the changes are expected to improve the overall performance of the application.

### DIFF
--- a/instrumentation_sql_test.go
+++ b/instrumentation_sql_test.go
@@ -232,6 +232,84 @@ func TestOpenSQLDB(t *testing.T) {
 	})
 }
 
+func TestPostgresDB(t *testing.T) {
+
+	recorder := instana.NewTestRecorder()
+	s := instana.NewSensorWithTracer(instana.NewTracerWithEverything(&instana.Options{
+		Service:     "go-sensor-test",
+		AgentClient: alwaysReadyClient{},
+	}, recorder))
+	defer instana.ShutdownSensor()
+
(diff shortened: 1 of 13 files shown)


instrumentation_sql_test.go
```diff
@@ -232,6 +232,84 @@ func TestOpenSQLDB(t *testing.T) {
 	})
 }
```
sanojsubran, line 234: @Angith Does it make sense to move these tests to a table driven test model?

Angith: Makes sense. I will do it @sanojsubran

Angith: Done 👍

sanojsubran (approved):

nithinputhenveettil (approved):
Good work @Angith 🎉

PR #943, opened 2024-11-05
Parsing gateway ip with bit shift instead of string on a loop

willianpc:
This is a small improvement at parsing the Gateway IP from /proc/net/route.
This parsing uses no loop and works by shifting bits in a number instead of parsing string for each octet of the IP.


willianpc:
Folks, feel free to ignore the PR if you don't see much value on it.
This is not a crucial part of the tracer and it just works fine today :)

sanojsubran:
Thanks @willianpc for the PR. It looks good. We will test and integrate it. Our team capacity is less this week. We will try to integrate it by next week at least.

sanojsubran (approved):

nithinputhenveettil (approved):
Thanks @willianpc, changes looks good to me.

PR #955, opened 2024-11-26
perf: optimize SQL based database instrumentation

Angith:
This PR consists of the following changes

1. removed the usage of regular expressions which imposes performance overhead on SQL span creation.
2. optimized the creation of SQL span tags by decoupling tag generation from query execution. Before, all span tags—including connection details, which are constants—were generated during query execution. This change moves the creation of these constant tags to the connection setup phase, allowing them to be reused across multiple queries executed through the same connection instance.
3. Span creation is now handled by a single object, enabling easier extension and the incorporation of more data in the future.
4. A `Tags` interface has been introduced to streamline the process of adding tags to spans, ensuring efficient tag management. This refactor also aims to improve code structure, enabling cleaner design and facilitating future expansion.


sql_wrappers.go
```diff
@@ -694,23 +694,23 @@ func get_conn_Execer_ExecerContext_Queryer_QueryerContext_ConnPrepareContext_Nam
 			sensor:      sensor,
 		},
 		Execer: &wExecer{
-			Execer:      Execer,
```
sanojsubran, line 697: @Angith We generate this file from tpl file in internal/bin/sql folder. Are we editing it manually?

Angith: Hi @sanojsubran. I edited this file manually. It seems it should not be edited manually. Let me try to generate it conventionally.

Angith: @sanojsubran I updated the `sql_wrappers.tpl` file to include the `sqlSpan` struct and generated the `sql_wrappers.go` using the updated template.

nithinputhenveettil (approved):
Looks good!

:shipit:

PR #964, opened 2024-12-03
fix: the test database is not cleaned up from the Cosmos account when integration tests fail

Angith:
The test database was not getting deleted while following cases
1. if the test fails by exceeding the total throughput of the Cosmos account. 
2. if the test panics.
3. If the user interrupts the test. 

The PR tries to tear down all the test data created for the above cases. 

**Exception**: Cleaning up is not feasible for the panic errors that happen during `m.Run()` in `TestMain`. Check this issue for more details: https://github.com/golang/go/issues/37206

### DIFF
--- a/instrumentation/instacosmos/container_client_test.go
+++ b/instrumentation/instacosmos/container_client_test.go
@@ -12,7 +12,9 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
@@ -65,6 +67,9 @@ var (
 	rec               *instana.Recorder
 )
 
+// exit error code
+var exitNotOk int = 1
+
 type alwaysReadyClient struct{}
 
 func (alwaysReadyClient) Ready() bool                                       { return true }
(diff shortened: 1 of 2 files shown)


instrumentation/instacosmos/container_client_test.go
```diff
@@ -65,6 +67,9 @@ var (
 	rec               *instana.Recorder
 )
 
+// exit error code
+var exitNotOk int = 1
```
nithinputhenveettil, line 71: Better to define as a `const`?

instrumentation/instacosmos/container_client_test.go
```diff
@@ -168,6 +194,18 @@ func TestMain(m *testing.M) {
 	tracer := instana.NewTracerWithEverything(&instana.Options{AgentClient: alwaysReadyClient{}}, recorder)
 	sensor := instana.NewSensorWithTracer(tracer)
 
+	// handles panic errors
+	// this will only work for the panic errors before m.Run()
+	// check this issue for more details: https://github.com/golang/go/issues/37206
+	defer func() {
+		if err := recover(); err != nil {
+			shutdown(sensor, exitNotOk)
```
nithinputhenveettil, line 202: Isn't it better to re-throw the panic after calling shutdown ?

nithinputhenveettil (changes requested):

instrumentation/instacosmos/container_client_test.go
Angith: done 👍

instrumentation/instacosmos/container_client_test.go
Angith: yes. Good Catch 👍

nithinputhenveettil (approved):
:shipit:

sanojsubran (approved):

PR #1012, opened 2025-01-14
fix: unit test failures while using instana.InitCollector() API

Angith:


collector.go
```diff
@@ -25,38 +26,50 @@ type Collector struct {
...
-		t:             tracer,
-		LeveledLogger: defaultLogger,
-		Sensor:        NewSensorWithTracer(tracer),
-	}
+	return c
+}
 
-	return C
+// GetC return the instance of instana Collector
+func GetC() TracerLogger {
```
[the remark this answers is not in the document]
nithinputhenveettil: Also I think we should handle the case where C is not initialised within the function. This way, users of the exported function will have clearer feedback and understanding of its state.

func GetC() (TracerLogger, error) {

collector.go
```diff
@@ -25,38 +26,50 @@ type Collector struct {
...
-		recorder: opts.Recorder,
-	}
+	})
 
-	C = &Collector{
-		t:             tracer,
-		LeveledLogger: defaultLogger,
-		Sensor:        NewSensorWithTracer(tracer),
-	}
+	return c
```
nithinputhenveettil, line 65: Is there a possibility where multiple goroutines might call `InitCollector` simultaneously, and while one of them is using `sync.Once`, another goroutine might return `c`, which could still be `nil` at that point? Should we implement some locking mechanism to prevent this scenario?

collector.go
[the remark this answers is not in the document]
Angith: Makes sense. Good suggestion @nithinputhenveettil 👍

collector.go
Angith: As we discussed directly, this is not a possible case. Each concurrent call waits for `sync.Once` to be finished and returns a non-nil value. Ensured this with a sample program. Hence marking the conversation resolved.

collector.go
```diff
@@ -25,38 +27,55 @@ type Collector struct {
...
+func GetCollector() (TracerLogger, error) {
+	muc.Lock()
+	defer muc.Unlock()
 
-	C = &Collector{
-		t:             tracer,
-		LeveledLogger: defaultLogger,
-		Sensor:        NewSensorWithTracer(tracer),
+	if _, ok := c.(*Collector); !ok {
+		return c, fmt.Errorf("collector is not initialized")
```
nithinputhenveettil, line 75:
```suggestion
		return c, fmt.Errorf("Collector has not been initialised yet. Please use InitCollector first")
```

nithinputhenveettil (approved):
Just added a small suggestion. 

Looks good, thank you for the refactor, @Angith.

collector.go
```diff
@@ -25,38 +27,55 @@ type Collector struct {
...
+func GetCollector() (TracerLogger, error) {
+	muc.Lock()
+	defer muc.Unlock()
 
-	C = &Collector{
-		t:             tracer,
-		LeveledLogger: defaultLogger,
-		Sensor:        NewSensorWithTracer(tracer),
+	if _, ok := c.(*Collector); !ok {
+		return c, fmt.Errorf("Collector has not been initialised yet. Please use InitCollector first")
```
sanojsubran, line 75:
```suggestion
		return c, fmt.Errorf("collector has not been initialised yet. Please use InitCollector first")
```
Error strings should begin with lowercase characters. Ref: https://go.dev/wiki/CodeReviewComments#error-strings

collector_test.go
```diff
@@ -15,23 +15,28 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func getInstanaCollector() instana.TracerLogger {
```
sanojsubran, line 18: Is this function really needed?

collector_test.go
```diff
@@ -15,23 +15,28 @@ import (
...
 
 func Test_Collector_LegacySensor(t *testing.T) {
 	recorder := instana.NewTestRecorder()
 	c := instana.InitCollector(&instana.Options{AgentClient: alwaysReadyClient{}, Recorder: recorder})
 	s := c.LegacySensor()
-	defer instana.ShutdownSensor()
+	defer instana.ShutdownCollector()
 
-	assert.NotNil(t, instana.C.LegacySensor())
+	assert.NotNil(t, getInstanaCollector().LegacySensor())
```
sanojsubran, line 39:
```suggestion
	assert.NotNil(t, c.LegacySensor())
```
I think it is better to remove the helper function altogether and use the instance directly. What do you think?

collector.go
```diff
@@ -25,38 +27,55 @@ type Collector struct {
...
-		opts.Recorder = NewRecorder()
-	}
+		StartMetrics(opts)
+
+		tracer := &tracerS{
+			recorder: opts.Recorder,
+		}
 
-	StartMetrics(opts)
+		muc.Lock()
```
sanojsubran, line 55: Query: Do we need this lock here?

sanojsubran (changes requested):
@Angith I have suggested some changes. Kindly let me know of your opinion.

collector_test.go
Angith: Not necessarily. Removed 👍

collector.go
Angith: Good catch. It is redundant.

sanojsubran (changes requested):

sanojsubran (approved):

PR #1023, opened 2025-01-27
rfactor: update Go Tracer unit tests with InitCollector API

Angith:
Update all unit tests with InitCollector() API replacing deprecated APIs.


adapters_test.go
```diff
@@ -17,13 +17,16 @@ import (
 
 func TestWithTracingSpan(t *testing.T) {
 	recorder := instana.NewTestRecorder()
-	s := instana.NewSensorWithTracer(instana.NewTracerWithEverything(&instana.Options{AgentClient: alwaysReadyClient{}}, recorder))
-	defer instana.ShutdownSensor()
+	s := instana.InitCollector(&instana.Options{
```
sanojsubran, line 20: Can we find another name for this variable?

azure_agent_test.go
```diff
@@ -69,9 +68,8 @@ func TestIntegration_AzureAgent_SendSpans(t *testing.T) {
 func TestIntegration_AzureAgent_SpanDetails(t *testing.T) {
 	defer agent.Reset()
 
-	tracer := instana.NewTracer()
-	sensor := instana.NewSensorWithTracer(tracer)
-	defer instana.ShutdownSensor()
+	sensor := instana.InitCollector(instana.DefaultOptions())
```
sanojsubran, line 71: Since the `Collector` has a `Sensor` as its member, can we rename this variable in all the testcases?

sanojsubran (changes requested):
@Angith The changes look good! I would like to suggest to retain the consistency in the naming of variables.

nithinputhenveettil (approved):

adapters_test.go
Angith: For consistency, I have made the variable name `c` in all unit tests.

azure_agent_test.go
Angith: For consistency, I have made the variable name `c` in all unit tests.

sanojsubran (approved):

PR #1040, opened 2025-02-11
refactor: updated pgx v5 instrumentation to support InitCollector API

Angith:


sanojsubran:
Good job @Angith ! Thanks for taking up this task and updating all the test cases!

PR #1104, opened 2025-04-22
fix: updated the dependencies of go-sensor

sanojsubran:
This PR updates the dependent packages to the latest version.


instrumentation/instaamqp/go.mod
```diff
@@ -1,17 +1,18 @@
 module github.com/instana/go-sensor/instrumentation/instaamqp
 
-go 1.22
+go 1.23
```
nithinputhenveettil, line 3: @sanojsubran : Do we need to update this? instaamqp is already deprecated, right ?

sanojsubran: @nithinputhenveettil : I'll revert it. But I think we can remove the package altogether after discussion.

nithinputhenveettil (approved):

PR #1110, opened 2025-04-25
fix: fixed tracer config precedence to the correct order

nithinputhenveettil:
Environment variables will have the highest priority. If they are not set, the values configured in the code will be used. If those are also not present, the configuration received from the agent will be applied. And if none of these are available, default values will be used.

### DIFF
--- a/fsm.go
+++ b/fsm.go
@@ -200,7 +200,10 @@ func (r *fsmS) handleRetries(e *f.Event, cb func(_ context.Context, e *f.Event),
 func (r *fsmS) applyHostAgentSettings(resp agentResponse) {
 	r.agentComm.from = newHostAgentFromS(int(resp.Pid), resp.HostID)
 
-	if resp.Secrets.Matcher != "" {
+	r.logger.Debug("agentOverrideSecrets flag value:", sensor.options.Tracer.agentOverrideSecrets)
+	r.logger.Debug("agent response received for secret config:", resp.Secrets.Matcher, resp.Secrets.List)
+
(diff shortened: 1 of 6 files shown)


Angith (approved):

options.go
```diff
@@ -98,7 +98,12 @@ func (opts *Options) setDefaults() {
 	}
 
 	if secretsMatcher == nil {
+		// If secretMatcher is nil, it means no in-code secret matcher has been configured,
+		// and the INSTANA_SECRETS environment variable also doesn't have a valid matcher.
+		// So, we will set a default secret matcher here and mark overrideAgent as true,
```
sanojsubran, line 103:
```suggestion
		// So, we will set a default secret matcher here and mark tracerDefaultSecrets as true,
```

sanojsubran (changes requested):
Done 👍

fsm.go
```diff
@@ -197,21 +197,40 @@ func (r *fsmS) handleRetries(e *f.Event, cb func(_ context.Context, e *f.Event),
...
 }
 
+func (r *fsmS) checkAndApplyHostAgentSecrets(resp agentResponse) error {
+	if !isTracerDefaultSecretsSet(sensor.options.Tracer) {
+		r.logger.Info("identified custom defined secrets matcher. Ignoring host agent default secrets configuration.")
+		return nil
+	}
+
+	if !isAgentSecretConfigValid(&resp) {
+
```
sanojsubran, line 207:
```suggestion
```

PR #1124, opened 2025-05-13
CI workflow to detect major version changes in instrumented packages

sanojsubran:
This PR helps to identify major version changes for the instrumented packages. 

In the Go ecosystem, when a new major version of a package is released, its module path changes, effectively making it a new package. Since we track new releases based on the module path, our CI pipeline currently cannot detect major version updates. This PR addresses that limitation by checking the GitHub releases of the package and notifying us when any new release, including major versions, is published.


nithinputhenveettil (approved):
All looks good to me, except for a few minor commented lines which I think can be removed. Good job, @sanojsubran

PR #1133, opened 2025-05-20
fix: instagorm unit test failure in windows

Angith:
### Error 
SQLite DB file is still open at the end of the test when it is to be deleted. On Windows, a file can not be deleted if another process or goroutine still has it open/locked, unlike on Unix where deletion can happen even while in use.


sanojsubran (approved):

nithinputhenveettil (commented):
@Angith : Some examples are using Instagorm. It’s better to update those as well, so that if a customer tries to run these examples on Windows, they get a working one.

nithinputhenveettil (approved):

PR #1164, opened 2025-06-16
refactor: optimised trace header extraction by adding an early exit when both headers are found

nithinputhenveettil:
refactor: optimised trace header extraction by adding an early exit when both headers are found

### DIFF
--- a/w3ctrace/context.go
+++ b/w3ctrace/context.go
@@ -52,18 +52,8 @@ func (c Context) IsZero() bool {
 func Extract(headers http.Header) (Context, error) {
 	var tr Context
 
-	for k, v := range headers {
-		if len(v) == 0 {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


w3ctrace/context.go
```diff
@@ -52,18 +52,8 @@ func (c Context) IsZero() bool {
 func Extract(headers http.Header) (Context, error) {
 	var tr Context
 
-	for k, v := range headers {
-		if len(v) == 0 {
-			continue
-		}
-
-		switch {
-		case strings.EqualFold(k, TraceParentHeader):
```
nithinputhenveettil, line 61: Since `header.Get` is already caseinsensitive, there is no need to use `strings.EqualFold` here.

sanojsubran (approved):

PR #1186, opened 2025-07-17
fix! : Ensure consistent default for MaxLogsPerSpan

Angith:
The logic of `MaxLogsPerSpan` is
- If MaxLogsPerSpan is set to 0, we can append as many logs.
- Else if MaxLogsPerSpan is set to a number, it will only add logs less than MaxLogsPerSpan
- If the user doesn't set Tracer Options with MaxLogsPerSpan while initializing the collector, MaxLogsPerSpan will be 0.
- If the user sets DefaultTracerOptions, MaxLogsPerSpan will be 2.

This PR ensure a consistent default for MaxLogsPerSpan across all code paths


sanojsubran (approved):

options.go
```diff
@@ -108,6 +108,10 @@ func (opts *Options) setDefaults() {
 
 	opts.Tracer.Secrets = secretsMatcher
 
+	if opts.Tracer.MaxLogsPerSpan == 0 {
```
sanojsubran, line 111: `setDefaults` should only set it to default values. No need to check. This will be problematic, if the user set it to zero.

sanojsubran (changes requested):
Please check my comments

options.go
Angith: We can not remove the check from the setDefaults function as it is called in initSensor function without any condition, and this will set the default MaxLogsPerSpan value always, even if user configured a custom value.

sanojsubran (approved):
Looks good. Good job @Angith !

tracer_test.go
```diff
@@ -89,3 +91,100 @@ func TestTracer_NonInstanaSpan(t *testing.T) {
...
+			defer instana.ShutdownSensor()
+
+			sp := tracer.StartSpan("test")
+			for _, field := range tC.args.fields {
+				sp.LogFields(field)
+			}
+			sp.Finish()
+
+			spans := recorder.GetQueuedSpans()
+			assert.Equal(t, tC.ExpectedSpanCount, len(spans))
```
nithinputhenveettil, line 187: Instead of only checking the overall span count, I think its better to examine each span individually, this will give us a clarity on which spans are retained and which ones are dropped due to the `MaxLogsPerSpan `config.

collector_test.go
```diff
@@ -140,6 +139,64 @@ func Test_Collector_Logger(t *testing.T) {
 	assert.Equal(t, 2, l.counter["error"])
 }
 
+func Test_Collector_MaxLogsPerSpan(t *testing.T) {
```
nithinputhenveettil, line 142:
```suggestion
func TestInitCollector_MaxLogsPerSpan(t *testing.T) {
```

tracer_options_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 func TestDefaultTracerOptions(t *testing.T) {
 	assert.Equal(t, instana.TracerOptions{
-		MaxLogsPerSpan: 2,
+		MaxLogsPerSpan: instana.MaxLogsPerSpan,
```
nithinputhenveettil, line 15: Suggestion : 

I think we can keep this as a hard coded `2`. If `MaxLogsPerSpan` is changed in the constants later, having it hard-coded here will make the test more meaningful. It implicitly conveys what the default value is expected to be, rather than just pulling it from the constant.

nithinputhenveettil (changes requested):
@Angith : Added a few comments.

Also, kindly update and add a more meaningful code comment for `MaxLogsPerSpan` in the following file. The current comment still refers to the old implementation logic. It would be helpful to add a clear explanation so that users understand what this option actually does.

https://github.com/instana/go-sensor/blob/6a5db5b61a13a549420025282321a85ad4b388ad/tracer_options.go#L10-L13


Thanks.

tracer_test.go
Angith: Makes sense. I have updated the unit tests.

PR #1254, opened 2025-09-12
feat: Ability to disable log collection at the tracer level

Angith:
This PR enables Go Tracer to disable log spans at tracer level.

There are four ways to disable log spans:
1. Using Code
2. Using Environment Variables
3. Using Configuration File
4. Using agent configuration


sanojsubran (changes requested):
@Angith I have done one round of review. Kindly check my comments. We can discuss together.

options.go
```diff
@@ -116,5 +120,97 @@ func (opts *Options) setDefaults() {
...
 }
+
+// parseInstanaTracingDisable processes the INSTANA_TRACING_DISABLE environment variable value
+// and updates the TracerOptions.Disable map accordingly.
+//
+// When the value is a boolean (true/false), the whole tracing feature is disabled/enabled.
+// When a list of category or type names is specified, only those will be disabled.
+//
+// Examples:
+// INSTANA_TRACING_DISABLE=True - disables all tracing
```
sanojsubran, line 143:
```suggestion
// INSTANA_TRACING_DISABLE=true - disables all tracing
```

options.go
```diff
@@ -116,5 +120,97 @@ func (opts *Options) setDefaults() {
...
+
+// parseConfigFile reads and parses the YAML configuration file at the given path
+// and updates the TracerOptions accordingly.
+//
+// The YAML file must follow this format:
+// tracing:
+//   disable:
+//     - logging: true
+
+func parseConfigFile(path string, opts *TracerOptions) error {
```
sanojsubran, line 184: Consider the security aspects since we are trying to read a file from the filesystem.

span.go
```diff
@@ -215,8 +227,10 @@ func (r *spanS) Tracer() ot.Tracer {
 // sendOpenTracingLogRecords converts OpenTracing log records that contain errors
 // to Instana log spans and sends them to the agent
 func (r *spanS) sendOpenTracingLogRecords() {
-	for _, lr := range r.Logs {
-		r.sendOpenTracingLogRecord(lr)
+	if !logging.disabled() {
+		for _, lr := range r.Logs {
+			r.sendOpenTracingLogRecord(lr)
+		}
```
sanojsubran, line 233:
```suggestion
	if logging.disabled() {
	    return
	}
	
	for _, lr := range r.Logs {
		r.sendOpenTracingLogRecord(lr)
	}
```

options.go
```diff
@@ -116,5 +121,149 @@ func (opts *Options) setDefaults() {
...
+	// Initialize the Disable map if it doesn't exist
+	if opts.DisableSpans == nil {
+		opts.DisableSpans = make(map[string]bool)
+	}
+
+	// Trim spaces from the value
+	value = strings.TrimSpace(value)
+
+	// if it's a boolean value, disable all categories
+	if isBooleanTrue(value) {
```
nithinputhenveettil, line 156:
```suggestion
	if strings.EqualFold(value, "true") {
```

options.go
```diff
@@ -116,5 +121,149 @@ func (opts *Options) setDefaults() {
...
+			opts.DisableSpans[item] = true
+		}
+	}
+}
+
+// isBooleanTrue checks if a string represents a boolean true value
+func isBooleanTrue(value string) bool {
+	value = strings.ToLower(value)
+	return value == "true"
+}
```
nithinputhenveettil, line 175:
```suggestion
```

options.go
```diff
@@ -116,5 +121,149 @@ func (opts *Options) setDefaults() {
...
+	return nil
+}
+
+// validateFile ensures the given config file path is safe and usable.
+// Security considerations:
+// - Resolves symlinks to prevent symlink attacks
+// - Ensures the path exists and is a regular file
+// - Enforces a reasonable file size limit to avoid DoS
+// - Warns if file permissions are too permissive (world-readable)
+func validateFile(path string) (absPath string, err error) {
```
nithinputhenveettil, line 232: I noticed that absPath is being returned along with the error in all the error cases. However, in situations where `filepath.Abs` or `EvalSymlinks` fails, this `absPath` itself may not be valid. This somewhat defeats the purpose of returning it along with the error.

To make debugging easier, you can consider:
– Returning the original `path` along with the error instead of `absPath`, since that is always the input value, or
– Returning both `path` and `absPath` for full context, so the caller can log both and know exactly where it failed.

This will ensure we are not passing a potentially invalid value back while still giving enough context for debugging.

tracer_options.go
```diff
@@ -22,6 +22,13 @@ type TracerOptions struct {
 	// See https://www.instana.com/docs/setup_and_manage/host_agent/configuration/#capture-custom-http-headers for details
 	CollectableHTTPHeaders []string
 
+	// Disable allows the exclusion of specific traces or calls from tracing based on the category (technology)
+	// or type (frameworks, libraries, instrumentations) supported by the traces.
+	// The main benefit of disabling is reducing the overall amount of data being collected and processed.
+
+	// See https://github.ibm.com/instana/technical-documentation/blob/master/tracing/specification/README.md#span-disabling for details
```
nithinputhenveettil, line 29:
```suggestion
```

IMO – This document is internal, so there’s no need to link it here.

nithinputhenveettil (changes requested):
@Angith : Just a few minor changes/suggestions.

docs/disabling_spans.md
```diff
@@ -0,0 +1,86 @@
...
+```yaml
+com.instana.tracing:
+  disable:
+    - logging
+```
+3. Restart the Instana agent:
+
+## Priority Order
+
+When multiple configuration methods are used, they are applied in the following order of precedence:
```
sanojsubran, line 74:
```suggestion
When multiple configuration methods are used, they are applied in the following decreasing order of precedence:
```

nithinputhenveettil (approved):
:shipit:

sanojsubran (approved):
Good job @Angith!

PR #1304, opened 2025-11-17
add support oracle tns parser for DbConnDetails

dantedenis:
9 We may need a parser for OracleTNS. Should we add this?
 
And it's not very safe to just save the URI in Raw, because it may contain secrets.
`instrumentation_sql.go 382:	  return DbConnDetails{RawString: connStr}
`
 It might be worth adding an optional parameter for custom escaping.

Alternatively, just add a global secret replacer that I can override in the project, and the tool will automatically check for its presence and execute the required action if it exists.

```go
var ReplaserSecretConnStr *regexp.Regexp
var ReplaserPattern = "***"

/*
...
*/

func ParseDBConnDetails(connStr string) DbConnDetails {
	strategies := [...]func(string) (DbConnDetails, bool){
		parseMySQLGoSQLDriver,
		parsePostgresConnDetailsKV,
		parseMySQLConnDetailsKV,
		parseRedisConnString,
		parseDBConnDetailsURI,
	}
	for _, parseFn := range strategies {
		if details, ok := parseFn(connStr); ok {
			return details
		}
	}

	d := DbConnDetails{RawString: connStr}
	if ReplaserSecretConnStr != nil {
		d.RawString = ReplaserSecretConnStr.ReplaceAllString(connStr, ReplaserPattern)
	}

	return d
}
```

Not explicitly, but we can control it.

Thx 😊


Angith:
Hi @dantedenis @vtolstov 
We have decided to move forward with releasing this PR with minor updates at this stage. Specifically, we will mask the password using `<redacted>` instead of `***`, as this is the convention commonly used in Instana. Please refer to the attached screenshot for a visual representation of the updated masking behavior.

ch incianro



The generic database span masking feature will be released in a future update. We will ensure that the existing Oracle DB span masking behavior remains consistent and is not impacted when the generic masking enhancement is released.

dantedenis:
Thank you for your cooperation

PR #1321, opened 2025-12-05
feat: add key-value format support for Oracle connections with masking password

Angith:


docs/sql.md
```diff
@@ -86,6 +86,32 @@ Instrumenting Db2 is the same as in the above example. You only need to pass the
 
 Ensure you use `go_ibm_db` as the driver name to correctly identify the IBM Db2 database.
 
+## Tracing Oracle Database
```
nithinputhenveettil: @Angith: How about adding a note here for Oracle, mentioning that the span format will be a generic SQL span?

PR #1377, opened 2026-02-05
feat: instrument echo v5

sanojsubran:
This PR adds support for the instrumentation of echo v5.

### DIFF
--- a/instrumentation/instaecho/v2/handler.go
+++ b/instrumentation/instaecho/v2/handler.go
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: MIT
 
+// Package instaechov2 provides instrumentation for the Echo v5 package
 package instaechov2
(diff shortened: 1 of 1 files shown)


instrumentation/instaecho/v2/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/instana/go-sensor/instrumentation/instaecho/v2
+
+go 1.25.0
+
+require (
+	github.com/instana/go-sensor v1.72.0
+	github.com/labstack/echo/v5 v5.0.1
```
nithinputhenveettil: @sanojsubran : Latest version is v5.0.4, it would be better if we could update this to the latest version.

sanojsubran: Done.

PR #1463, opened 2026-04-20
feat: add poll rate support for metrics

Angith:


options.go
```diff
@@ -123,6 +126,43 @@ func (opts *Options) applyProfilingConfiguration() {
...
+			defaultLogger.Warn("INSTANA_METRICS_TRANSMISSION_DELAY must be positive, using default 1000ms")
+			opts.Metrics.TransmissionDelay = 1000
+		} else {
+			// Valid value from ENV
+			opts.Metrics.TransmissionDelay = delay
+		}
+	}
+
+	// Step 3: Enforce minimum value (1000ms)
+	if opts.Metrics.TransmissionDelay < 1000 {
```
sanojsubran: Use constants for min, max values

options.go
```diff
@@ -123,6 +126,43 @@ func (opts *Options) applyProfilingConfiguration() {
...
+		defaultLogger.Warn("metrics transmission delay is below minimum 1000ms, using minimum 1000ms (configured: ", opts.Metrics.TransmissionDelay, "ms)")
+		opts.Metrics.TransmissionDelay = 1000
+	}
+
+	// Step 4: Enforce maximum cap (5000ms)
+	if opts.Metrics.TransmissionDelay > 5000 {
```
sanojsubran: Use constants for min, max values

README.md
```diff
@@ -66,6 +66,34 @@ func init() {
...
+col = instana.InitCollector(&instana.Options{
+  Service: "My app",
+  Metrics: instana.MetricsOptions{
+    TransmissionDelay: 3000,  // 3 seconds
+  },
+})
+```
+
+**Configuration Rules:**
+- Valid range: 1000ms to 5000ms
```
sanojsubran, line 90: Please check whether the `pollrate` is agent config is in `ms` or `s`. Let's follow the same unit at the tracer side also

meter.go
```diff
@@ -62,6 +87,16 @@ func (m *meterS) Stop() {
 	m.done <- struct{}{}
 }
 
+func getTransmissionDelay(options *Options) time.Duration {
+	interval := time.Duration(options.Metrics.TransmissionDelay) * time.Millisecond
+	// Safety check: fallback to default if interval becomes negative,
```
sanojsubran, line 98: Suggestion: Return an error/warning also here and at the callee side, log a message based on the error state. We need to know whether this safety check got triggered or not

sanojsubran (dismissed):

fsm.go
```diff
@@ -268,10 +268,30 @@ func (r *fsmS) applyHostAgentSettings(resp agentResponse) {
...
+	// Check if sensor is initialized
+	if sensor == nil || sensor.options == nil {
+		r.logger.Debug("Sensor not initialized, skipping poll_rate configuration")
+		return
+	}
+
+	// If no poll rate is provided by agent, use default (1 second)
+	if resp.PluginConfig.PollRate == 0 {
+		r.logger.Debug("No poll_rate configuration received from agent, using default 1 second")
+		sensor.options.Metrics.setTransmissionInterval(1)
```
sanojsubran: Remove the hardcoding of 1

fsm.go
```diff
@@ -420,6 +440,7 @@ func (r *fsmS) reset() {
 
 func (r *fsmS) ready(_ context.Context, e *f.Event) {
 	go delayed.flush()
+	sensor.meter.Run(sensor.options.Metrics.getTransmissionInterval())
```
sanojsubran: Ensure that the sensor will be initialized properly before calling this method

meter.go
```diff
@@ -23,43 +32,105 @@ type MetricsS acceptor.Metrics
...
+}
+
+// getTransmissionInterval returns the current metrics transmission interval.
+// This value is configured through the agent's configuration.yaml file.
+func (m *MetricsOptions) getTransmissionInterval() time.Duration {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	if m.transmissionInterval == 0 {
+		return defaultTransmissionInterval * time.Second
```
sanojsubran: Don't we have to update `m.transmissionInterval` if it is zero?

meter.go
```diff
@@ -23,43 +32,105 @@ type MetricsS acceptor.Metrics
...
+	}
+	return m.transmissionInterval
+}
+
+// setTransmissionInterval sets the metrics transmission interval.
+// This is an internal method called when agent configuration is received during
+// the initial handshake. Valid range: minTransmissionInterval-maxTransmissionInterval seconds.
+// Values outside the range are clamped to the nearest bound.
+func (m *MetricsOptions) setTransmissionInterval(seconds int) {
```
sanojsubran, line 60: You can avoid holding a lock while logging by computing the value first and only synchronizing for the assignment. Logging can block or take time, so it's generally better not to perform it while holding a mutex

sanojsubran (changes requested):

fsm.go
Angith: Good catch. Done 👍

meter.go
Angith: Makes sense. Done 👍

fsm.go
Angith: done 👍

meter.go
Angith: I think no update is needed here for the following reasons.

1. setTransmissionInterval is called during the agent handshake before meter.Run is ever invoked, so transmissionInterval is never zero by the time it matters in the normal (non-serverless) path.
2. Serverless agents skip the FSM entirely and call getTransmissionInterval() directly with the field at its zero value. The == 0 guard correctly returns 1s here.
3. Writing here would break the read lock contract. Mutating inside it is extra complexity with zero benefit.

Angith: As we discussed, I have removed this fallback logic from here.

sanojsubran (approved):

nithinputhenveettil (approved):

PR #1496, opened 2026-05-29
chore(ci): add Go compatibility tests for latest two supported streams

nithinputhenveettil:
- add Go compatibility tests for latest two supported streams
- add support for Go runtime support table to currency report


.github/workflows/go-support-matrix-tests.yml
```diff
@@ -0,0 +1,153 @@
...
+          echo "go_versions=$GO_VERSIONS" >> "$GITHUB_OUTPUT"
+
+  build:
+    name: Go ${{ matrix.go-version }} tests
+    needs: generate-matrix
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
sanojsubran, line 63: @nithinputhenveettil If we increase this, will it help it reducing the run time of the job?

Angith (approved):

sanojsubran (approved):

.github/workflows/go-support-matrix-tests.yml
Angith: @sanojsubran Yes, but running the tests in parallel may exhaust the throughput of the Azure Cosmos DB, which is currently configured at 3000 RU/s. @nithinputhenveettil right?

nithinputhenveettil: @sanojsubran: Yes, increasing the parallelism will help reduce the overall runtime. However, when we do that, the Cosmos integration tests sometimes fail with throughput errors.

If needed, we can experiment with increasing the parallelism and find the maximum level that works reliably. That said, it is difficult to guarantee a specific number that will never hit throughput issues.

Since this workflow runs only once a week on Sunday midnight, I don't think the execution time is a major concern. From my testing, the full run takes around 1 to 1.5 hours to complete.

PR #1510, opened 2026-06-09
fix: prevent duplicate meter initialization in InitSensor

Angith:


nithinputhenveettil (approved):

sanojsubran (approved):

sensor.go
```diff
@@ -213,6 +213,12 @@ func (r *sensorS) serviceOrBinaryName() string {
 //
 // Deprecated: Use [StartMetrics] instead.
 func InitSensor(options *Options) {
+	// If sensor is already initialized, return
+	s, _ := getSensor()
+	if s != nil {
+		return
```
nithinputhenveettil, line 220: @Angith : 

Please add a debug log here, something like:
```
Sensor is already initialised, so returning without reinitialising it.
```

